PR #1, opened 2023-02-09
Infrastructure

al1img:


Eabramov84 (approved):

src/iam/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+set(TARGET aosiamcpp)
+
+# ######################################################################################################################
+# Sources
+# ######################################################################################################################
+
+set(SOURCES certhandler/certhandler.cpp)
```
MykolaSolyanko, line 7: i think it could be better to use 
`file(GLOB_RECURSE SOURCES CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")`

in this case it will be easier to add new source files

src/sm/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+set(TARGET aossmcpp)
+
+# ######################################################################################################################
+# Sources
+# ######################################################################################################################
+
+set(SOURCES launcher/launcher.cpp)
```
MykolaSolyanko, line 7: ditto

PR #3, opened 2023-02-14
Log implementation

al1img:


src/common/utils/stringer.hpp
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Copyright (C) 2023 Renesas Electronics Corporation.
+// Copyright (C) 2023 EPAM Systems, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
Eabramov84: remove licence text

src/common/log/log.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+#include "config.hpp"
+#include "utils/stringer.hpp"
+#include "utils/utils.hpp"
+
+/**
+ * Helper macro to display debug log.
+ */
+#if CONFIG_LOG_LEVEL >= CONFIG_LOG_LEVEL_DEBUG
+#define LOG_MODULE_DBG(module) aos::Log(module, LogLevelType::eDebug)
```
MykolaSolyanko, line 22: Perhaps we could consider including information about the function using the `__FUNCTION__` macro?

src/common/log/log.hpp
```diff
@@ -0,0 +1,249 @@
...
+     * Logs object that implements Stringer interface.
+     *
+     * @param stringer object to log.
+     * @return Log&
+     */
+    Log& operator<<(const Stringer& stringer)
+    {
+        *this << stringer.ToString();
+
+        return *this;
```
MykolaSolyanko, line 187: or you can write 
`return operator<<(stringer.ToString());`

src/common/utils/stringer.hpp
```diff
@@ -0,0 +1,133 @@
...
+     * @return bool result.
+     */
+    friend bool operator!=(T type, const EnumStringer<T>& stringer) { return stringer.GetValue() != type; };
+
+    /**
+     * Converts enum value to string.
+     *
+     * @return string.
+     */
+    const char* ToString() const override
```
MykolaSolyanko, line 110: Could we consider returning a Pair consisting of an error and a string, which would allow for the possibility of checking conversions?

src/common/error/error.hpp
```diff
@@ -1,30 +1,58 @@
...
+     * @return bool result.
+     */
+    bool Is(const Error& e) const { return GetValue() == e.GetValue(); }
+
+    /**
+     * Returns string representation of error type.
+     *
+     * @return error string.
+     */
+    virtual size_t GetStrings(const char* const** strings) const override
```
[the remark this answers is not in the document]
al1img: We don't use STL.

src/common/log/log.hpp
```diff
@@ -0,0 +1,249 @@
...
+#include "utils/stringer.hpp"
+#include "utils/utils.hpp"
+
+/**
+ * Helper macro to display debug log.
+ */
+#if CONFIG_LOG_LEVEL >= CONFIG_LOG_LEVEL_DEBUG
+#define LOG_MODULE_DBG(module) aos::Log(module, LogLevelType::eDebug)
+#else
+#define LOG_MODULE_DBG(module) true ? (void)0 : aos::LogVoid() & aos::Log(module, LogLevelType::eDebug)
```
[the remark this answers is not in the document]
al1img: It is a trick. To make log line not compiled at all. With (void)0 it is not possible because `(void)0 << "some string";` is not compiled but: `true ? (void)0 : aos::LogVoid() & aos::Log(module, LogLevelType::eDebug) << "some string";` compiles and not included to the final binary.

src/common/error/error.hpp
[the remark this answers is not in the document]
MykolaSolyanko: i don't mean to use STL this type issimple to write
```
template <typename T1, typename T2>
struct Pair {
  T1 first; 
  T2 second;
};

File: src/common/log/log.hpp (hunk 2 above)
Comment (al1img) [reply src/common/log/log.hpp:22]: I guess it is overhead. The developer always can find the line by log payload. For others this information is useless. Would be useful in assert.

Comment (MykolaSolyanko) [reply src/common/log/log.hpp:22]: It's up to you, but what I meant was the following use case: when you use `LOG_DBG() << "first" << 1 << "second"`, the output is `first 1 second`. If you try to find this string later, you won't be able to locate it easily. Therefore, it would be better to have a function that makes it easier to identify where the log is coming from.

File: src/common/utils/stringer.hpp (hunk 4 above)
Comment (MykolaSolyanko) [reply src/common/utils/stringer.hpp:110]: We need to check if the conversion to a string was successful or not, because if it fails, the log output will show undefined. This can be confusing and it's not clear what it means.

Comment (MykolaSolyanko) [reply src/common/utils/stringer.hpp:110]: This is just a suggestion, so it's entirely up to you.

File: src/common/error/error.hpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (al1img) [reply src/common/error/error.hpp:47]: accepted.

File: src/common/utils/stringer.hpp (hunk 4 above)
Comment (al1img) [reply src/common/utils/stringer.hpp:114]: Still do not clearly get the use case. `ToString()` method is used in the log for example like this:
```c++
LOG_DBG() << SomeStringerInstance();
```
Why do we need to check if conversion is successful or not in this case, and what we shall do if the conversion is not successful?

src/common/log/log.hpp
al1img: We have module at least for now. Let's go without function name for now. Can be added later if we experience problems.

MykolaSolyanko: ok, agreed

MykolaSolyanko (approved):

Eabramov84 (approved):

PR #6, opened 2023-02-23
Thread wrapper

al1img:


Eabramov84 (approved):

include/aos/common/thread.hpp
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Unlocks Aos mutex.
+     *
+     * @return Error.
+     */
+    Error Unlock() { return pthread_mutex_unlock(&mPMutex); }
+
+private:
+    friend class ConditionalVariable;
```
MykolaSolyanko, line 93: Additionally, you could define a conversion operator from Mutex to pthread_mutex_t, which would allow you to use the Mutex object directly inside the ConditionalVariable class without needing to declare it as a friend.

include/aos/common/thread.hpp
```diff
@@ -0,0 +1,290 @@
...
+     * Destroys lock guard instance.
+     */
+    ~LockGuard() { mMutex.Unlock(); }
+
+    /**
+     * Returns current lock guard error.
+     *
+     * @return Error&.
+     */
+    Error& GetError() { return mError; }
```
MykolaSolyanko, line 124: why return by references ? it could be potentially dangling references

include/aos/common/thread.hpp
```diff
@@ -11,13 +11,14 @@
...
 #include "aos/common/error.hpp"
+#include "aos/common/noncopyable.hpp"
 
 namespace aos {
 
 /**
  * Aos thread.
  */
-class Thread {
+class Thread : private NonCopyable {
```
MykolaSolyanko, line 21: I suggest adding a move constructor and move assignment operator to this class because the delete constructor has suppressed the default implementation of these operations.

MykolaSolyanko (approved):

include/aos/common/error.hpp
```diff
@@ -35,8 +39,23 @@ typedef ErrorType::Enum ErrorEnum;
...
+     */
     using EnumStringer<ErrorType>::EnumStringer;
 
+    // cppcheck-suppress noExplicitConstructor
+    /**
+     * Constructs error instance from system errno value.
+     *
+     * @param e errno.
+     */
+    Error(int e)
```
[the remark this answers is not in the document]
al1img: I don't want to add a default constructor. I would like the user explicitely specifies the error type.

include/aos/common/error.hpp
```diff
@@ -48,7 +67,31 @@ class Error : public EnumStringer<ErrorType> {
      * @param e error to check.
      * @return bool result.
      */
-    bool Is(const Error& e) const { return GetValue() == e.GetValue(); }
+    bool Is(const Error& e) const
```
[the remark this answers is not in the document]
al1img: Base class `EnumStringer<ErrorType>` has this operator.

include/aos/common/thread.hpp
```diff
@@ -0,0 +1,290 @@
...
+     */
+    Error Join() { return pthread_join(mPThread, nullptr); }
+
+private:
+    void* (*mAdapter)(void*);
+    void*     mArg;
+    pthread_t mPThread;
+
+    Thread(Thread const&) = delete;
+    Thread& operator=(Thread const&) = delete;
```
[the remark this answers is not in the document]
al1img: Moved to NonCopyable class.

include/aos/common/thread.hpp
al1img: We will not use move semantics for now.

include/aos/common/thread.hpp
al1img: ok

al1img: done

PR #7, opened 2023-02-28
Array and initial launcher interfaces

al1img:


include/aos/common/array.hpp
```diff
@@ -0,0 +1,368 @@
...
+    {
+        SetBuffer(buffer);
+    }
+
+    /**
+     * Creates array instance from const C array.
+     *
+     * @param items const C array.
+     */
+    explicit Array(const T* items)
```
MykolaSolyanko, line 43: maybe better to use range from start to end, and in this case it will be more safety ?
```
Array(const T* start, const T* end)
```
and calculate correct mSize

include/aos/common/array.hpp
```diff
@@ -0,0 +1,368 @@
...
+        , mMaxSize(cMaxSize)
+    {
+    }
+
+    /**
+     * Creates array instance from another array.
+     *
+     * @param array another array instance.
+     */
+    Array(const Array& array)
```
MykolaSolyanko, line 55: maybe add move constructor?

include/aos/common/array.hpp
```diff
@@ -0,0 +1,368 @@
...
+     */
+    T& operator[](size_t index) { return mItems[index]; }
+
+    /**
+     * Provides access to array const item by index.
+     *
+     * @param index item index.
+     * @return const T&.
+     */
+    const T& operator[](size_t index) const { return mItems[index]; }
```
MykolaSolyanko, line 161: if you use assert maybe add assert for check out_of_range ?

include/aos/common/array.hpp
```diff
@@ -0,0 +1,368 @@
...
+
+    /**
+     * Provides access to array item by index with boundaries check.
+     *
+     * @param index item index.
+     * @param item return item.
+     * @return Error.
+     */
+    Error At(size_t index, T** item)
```
MykolaSolyanko, line 170: what if return `Pair<Error, T*>` ?
I mean it will be helpful because it's not necessary to create variables and pass them to function

include/aos/common/array.hpp
```diff
@@ -0,0 +1,368 @@
...
+        return ErrorEnum::eNone;
+    }
+
+    /**
+     * Returns array first item.
+     *
+     * @param item return item.
+     * @return Error.
+     */
+    Error Front(T** item) { return At(0, item); }
```
MykolaSolyanko, line 209: ditto

include/aos/common/config/types.hpp
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#ifndef AOS_CONFIG_TYPES_HPP_
+#define AOS_CONFIG_TYPES_HPP_
+
+/**
+ * Service ID len.
+ */
+#ifndef CONFIG_SERVICE_ID_LEN
+#define CONFIG_SERVICE_ID_LEN 40
```
MykolaSolyanko: we can use consexpr instead define because constexpr is more modern C++ way

MykolaSolyanko (approved):

include/aos/common/array.hpp
MykolaSolyanko: You have the following constructor: `Array(const T* items)`. In your example, you used `Array<const int> array(sValue, ArraySize(sValue))`, but this constructor was not called. Instead, `Array(const T* items)` was called in the example `Array<const int> array(sValue)`. Therefore, we don't have information about the size of the passed array.

Eabramov84 (approved):

PR #9, opened 2023-03-08
Thread pool

al1img:


Eabramov84 (approved):

include/aos/common/thread.hpp
```diff
@@ -375,6 +386,158 @@ class ConditionalVariable : private NonCopyable {
...
+        static_assert(sizeof(Function<T>) < cMaxTaskSize, "task is bigger than max task size");
+
+        LockGuard lock(mMutex);
+
+        auto err = mBuffer.PushValue(Function<T>(functor, arg));
+        if (!err.IsNone()) {
+            return err;
+        }
+
+        if (!(err = mCondVar.NotifyAll()).IsNone()) {
```
MykolaSolyanko, line 430: You can use NotifyOne to wake up only one thread instead of all available

include/aos/common/thread.hpp
```diff
@@ -375,6 +386,158 @@ class ConditionalVariable : private NonCopyable {
...
+                    UniqueLock lock(mMutex);
+
+                    auto err = mCondVar.Wait([this]() { return mShutdown || !mBuffer.IsEmpty(); });
+                    assert(err.IsNone());
+
+                    if (mShutdown) {
+                        return;
+                    }
+
+                    if (mBuffer.IsEmpty()) {
```
MykolaSolyanko, line 462: In my opinion, this check is not necessary because when we exit from `Wait` condition, we are always holding the mutex. Therefore, if someone tries to call AddTask, they will be blocked.

include/aos/common/thread.hpp
```diff
@@ -375,6 +386,158 @@ class ConditionalVariable : private NonCopyable {
...
+
+                    lock.Unlock();
+
+                    (*static_cast<CallableItf*>(static_cast<void*>(taskBuffer)))();
+
+                    lock.Lock();
+
+                    mThreadsCount--;
+
+                    err = mCondVar.NotifyAll();
```
MykolaSolyanko, line 479: Maybe we should use a separate condition variable for the Wait method. This way, we would only wake up the threads that are waiting for a specific condition, instead of waking up all the threads and making them check their condition. Otherwise, all the threads that are woken up will compete for the mutex.

MykolaSolyanko (approved):

include/aos/common/thread.hpp
al1img: I didn't get this point. This condition prevents popping task from the empty buffer. Also, It is not by design that `AddTask` blocks. It either adds task to queue or returns error.

include/aos/common/thread.hpp
al1img: I can not, because it could be one waiting in `Wait` api.

include/aos/common/thread.hpp
al1img: Indeed, it is good proposal. Will redo.

MykolaSolyanko (approved):

PR #11, opened 2023-03-13
Add ServiceInfo and LayerInfo

al1img:


include/aos/common/error.hpp
```diff
@@ -72,6 +72,38 @@ class Error : public EnumStringer<ErrorType> {
...
+        }
+    }
+
+    /**
+     * Assigns error from another error.
+     *
+     * @param err error to copy from.
+     */
+    Error& operator=(const Error& err)
+    {
```
MykolaSolyanko, line 96: If we have a derived class, then in the copy constructor as well as in the copy assignment operator, we need to invoke the copy assignment operator for the base class. In this case, it would result in the subobject of the base class has an incorrect value.
```
EnumStringer::operator=(err.GetValue())
```

include/aos/common/types.hpp
```diff
@@ -36,6 +51,25 @@ constexpr auto cInstanceIDLen = AOS_CONFIG_TYPES_INSTANCE_ID_LEN;
...
+ */
+constexpr auto cVendorVersionLen = AOS_CONFIG_TYPES_VENDOR_VERSION_LEN;
+
+/**
+ * Service/layer description len.
+ */
+
+constexpr auto cDescriptionLen = AOS_CONFIG_TYPES_DESCRIPTION_LEN;
+
+constexpr auto cSHA256Size = 64;
```
MykolaSolyanko, line 70: Isn't the size of SHA256 32 bytes? Or is it intentionally specified differently?

MykolaSolyanko (approved):

include/aos/common/error.hpp
al1img: You are right. Fixed.

include/aos/common/types.hpp
al1img: Mindlessly copied from grpc options.Fixed.

MykolaSolyanko (approved):

PR #43, opened 2023-10-09
[common, iam] Add CertHandler interface.

mykola-kobets-epam:


include/aos/iam/certhandler.hpp
```diff
@@ -9,11 +9,117 @@
...
+#include "aos/common/tools/string.hpp"
+#include "aos/common/tools/thread.hpp"
+#include "aos/common/tools/log.hpp"
+#include "aos/common/tools/utils.hpp"
+#include "aos/common/types.hpp"
+
+// Missing types
+// Time
+// config::ModuleConfig => from config.go
+// ASN1, x509
```
al1img, line 23: Need to check how if there is lib to replace or implement ourselves.

include/aos/iam/certhandler.hpp
```diff
@@ -9,11 +9,117 @@
...
+struct CertInfo {
+	aos::Issuer Issuer;
+	CertificateSerialNum Serial;
+	Url CertURL;
+	Url KeyURL;
+	Time NotAfter;
+};
+
+// CertStorage provides API to store/retrieve certificates info.
+class CertStorage {
```
al1img, line 88: Rename to storage. All interfaces have `Itf` suffix.

include/aos/iam/certhandler.hpp
```diff
@@ -9,11 +9,117 @@
...
+public:
+	virtual Error AddCertificate(const CertificateType& certType, const CertInfo& cert) = 0;
+	virtual Error GetCertificate(const Issuer& issuer, const CertificateSerialNum& serial, CertInfo& res) = 0;
+	virtual Error GetCertificates(const CertificateType& certType, Array<CertInfo>& res) = 0;
+	virtual Error RemoveCertificate(const CertificateType& certType, const Url& certURL) = 0;
+	virtual Error RemoveAllCertificates(const CertificateType& certType) = 0;
+};
+
+// CertModule provides API to manage module certificates.
+class CertModule {
```
al1img, line 98: `Itf` suffix.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+// CertStorage provides API to store/retrieve certificates info.
+class CertStorageItf {
+public:
+    virtual Error AddCertificate(const String& certType, const CertInfo& cert) = 0;
+    virtual Error GetCertificate(const String& issuer, const String& serial, CertInfo& res) = 0;
+    virtual Error GetCertificates(const String& certType, Array<CertInfo>& res) = 0;
+    virtual Error RemoveCertificate(const String& certType, const String& certURL) = 0;
+    virtual Error RemoveAllCertificates(const String& certType) = 0;
+    virtual ~CertStorageItf() = default;
+};
```
al1img, line 89: In storage let's rename all methods as following:

`AddCertInfo`
`GetCertInfo`
etc.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+    virtual Error ValidateCertificates(
+        Array<CertInfo>& validInfo, Array<Url>& invalidCerts, Array<CertificatePublicKey>& invalidKeys)
+        = 0;
+    virtual Error                            SetOwner(const String& password) = 0;
+    virtual Error                            Clear() = 0;
+    virtual RetWithError<crypto::PrivateKey> CreateKey(const String& password, const String& algorithm) = 0;
+    virtual Error ApplyCertificate(const Array<x509::Certificate*>& certs, CertInfo& certInfo, Password& password) = 0;
+    virtual Error RemoveCertificate(const String& certURL, const String& password) = 0;
+    virtual Error RemoveKey(const String& certURL, const String& password) = 0;
+    virtual Error Close() = 0; // TODO: probably it is not required as we have destructors
```
al1img, line 103: Yes, no need for close.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+    virtual ~CertModuleItf() = default;
+};
+
+// moduleDescriptor
```
al1img, line 107: Remove it.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+// aoscloud/aos_common/utils/cryptutils
 
 namespace aos {
 namespace iam {
 namespace certhandler {
 
+/***********************************************************************************************************************
+ *
+ **********************************************************************************************************************/
+struct Time {
```
al1img, line 35: Create `common/time.hpp' and move this struct there.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+    static constexpr uint64_t Hour = 60 * 60 * 1000; // ms.
+};
+
+namespace crypto {
+using PrivateKey = String;
```
al1img, line 40: Create `common/crypto.hpp` and move `PrivateKey`, `Certificate` there.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+}
+
+namespace x509 {
+using Certificate = Array<uint8_t>;
+}
+
+namespace config {
```
al1img, line 47: Remove

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+struct ModuleConfig { };
+} // namespace config
+
+namespace json {
```
al1img, line 51: ditto

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+using RawMessage = String;
+}
+
+template <class... Args>
```
al1img, line 55: ditto

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+// CertModule provides API to manage module certificates.
+class CertModuleItf {
+public:
+    virtual Error ValidateCertificates(
+        Array<CertInfo>& validInfo, Array<Url>& invalidCerts, Array<CertificatePublicKey>& invalidKeys)
+        = 0;
+    virtual Error                            SetOwner(const String& password) = 0;
+    virtual Error                            Clear() = 0;
+    virtual RetWithError<crypto::PrivateKey> CreateKey(const String& password, const String& algorithm) = 0;
+    virtual Error ApplyCertificate(const Array<x509::Certificate*>& certs, CertInfo& certInfo, Password& password) = 0;
```
al1img, line 100: const CertInfo, const Password

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+    virtual Error RemoveCertificate(const String& certType, const String& certURL) = 0;
+    virtual Error RemoveAllCertificates(const String& certType) = 0;
+    virtual ~CertStorageItf() = default;
+};
+
+// CertModule provides API to manage module certificates.
+class CertModuleItf {
+public:
+    virtual Error ValidateCertificates(
+        Array<CertInfo>& validInfo, Array<Url>& invalidCerts, Array<CertificatePublicKey>& invalidKeys)
```
al1img, line 95: `invalidKeys` is also Url

include/aos/iam/certhandler.hpp
```diff
@@ -23,12 +121,34 @@ namespace certhandler {
...
+    Array<CertificateType> GetCertTypes();
+
+    // SetOwner owns security storage.
+    Error SetOwner(const String& certType, const String& password);
+
+    // Clear clears security storage.
+    Error Clear(const String& string);
+
+    // CreateKey creates key pair.
+    RetWithError<Array<uint8_t> /*csr*/> CreateKey(
```
al1img, line 137: CSR could be quite big, return with parameter.

include/aos/iam/certhandler.hpp
```diff
@@ -23,12 +121,34 @@ namespace certhandler {
...
+        const String& certType, const String& subject, const String& password);
+
+    // ApplyCertificate applies certificate.
+    Error ApplyCertificate(
```
al1img, line 141: Let's return `CertInfo` with RetWithError.

include/aos/iam/certhandler.hpp
```diff
@@ -23,12 +121,34 @@ namespace certhandler {
...
+        const String& certType, const Array<uint8_t>& cert, Url& certURL, CertificateSerialNum& serial);
+
+    // GetCertificate returns certificate info.
+    Error GetCertificate(
```
al1img, line 145: ditto

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,110 @@
...
+
+// CertModule provides API to manage module certificates.
+class CertModuleItf {
+public:
+    virtual Error ValidateCertificates(
+        Array<CertInfo>& validInfo, Array<Url>& invalidCerts, Array<CertificatePublicKey>& invalidKeys)
+        = 0;
+    virtual Error                            SetOwner(const String& password) = 0;
+    virtual Error                            Clear() = 0;
+    virtual RetWithError<crypto::PrivateKey> CreateKey(const String& password, const String& algorithm) = 0;
```
al1img, line 99: I think here will be reference to PrivateKey.

include/aos/iam/certhandler.hpp
```diff
@@ -24,11 +190,82 @@ namespace certhandler {
 class CertHandler {
 public:
     /**
-     * Creates key.
+     * RegisterModule registers module.
```
al1img, line 193: In go it was required to put method or var name in first place in comment. For oxygen it is not required. Should be just: `Registers module`. For other methods as well.

include/aos/iam/certhandler.hpp
```diff
@@ -24,11 +190,82 @@ namespace certhandler {
...
+     *
+     * @param certType certificate type.
+     * @param issuer issuer name.
+     * @param serial serial number.
+     * @returns RetWithError<CertInfo>.
+     */
+    RetWithError<CertInfo> GetCertificate(const String& certType, const Array<uint8_t>& issuer, const String& serial);
+
+    /**
+     * GetCertificate returns certificate info.
```
al1img, line 257: Wrong comment.

include/aos/iam/certhandler.hpp
```diff
@@ -18,37 +18,12 @@
...
+#include "aos/iam/certmodule.hpp"
 
 namespace aos {
 namespace iam {
 namespace certhandler {
 
-/**
- * General information information about certificate.
- */
-struct CertInfo {
```
al1img: Definitely, should be kept in this file.

include/aos/iam/certhandler.hpp
```diff
@@ -18,37 +18,12 @@
 #include "aos/common/tools/thread.hpp"
 #include "aos/common/tools/utils.hpp"
 #include "aos/common/types.hpp"
+#include "aos/iam/certmodule.hpp"
 
 namespace aos {
 namespace iam {
 namespace certhandler {
 
-/**
```
al1img: 1. In general, no need to move CertModuleItf into separate file. Usually, we define interfaces in the same header where the client of these interfaces is defined. Exceptions are generic interfaces used by many clients. In this particular case `CertModuleItf` could be placed in separate file but I would like to have clean strict rules. So, keep it in the `certhandler.hpp` now.
2. These changes are not related to this commit.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,205 @@
...
+constexpr auto cModuleParamsMaxLen = AOS_IAM_TYPES_MODULE_PARAMS_MAX_LEN;
+
+/**
+ * ModuleConfig module configuration.
+ */
+struct ModuleConfig {
+    /**
+     * Certificate signing algorithm.
+     */
+    StaticString<cCertificateSignAlgorithmLen> mSignatureAlgorithm;
```
al1img, line 45: Better to have enum for this.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,205 @@
...
+    StaticArray<StaticString<cModuleKeyUsageNameLen>, cModuleKeyUsagesMaxCount> mExtendedKeyUsage;
+    /**
+     * Alternative DNS names.
+     */
+    StaticArray<StaticString<cDnsNameLen>, cAltDnsNamesCount> mAlternativeNames;
+    /**
+     * TODO: probable not required anymore.
+     * Module Disabled flag.
+     */
+    bool mDisabled = 0;
```
al1img, line 62: Right not neede in our case.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Skip certificate chain validation.
+     */
+    bool mSkipValidation = 0;
+    /**
+     * TODO: didnt undertand why they are required from implementation.
+     */
+    StaticString<cModuleParamsMaxLen> mParams;
```
al1img, line 70: It keeps specific for different modules parameters. Not needed here.

include/aos/iam/certhandler.hpp
```diff
@@ -8,12 +8,183 @@
...
+     */
+    virtual Error ApplyCertificateChain(
+        const Array<crypto::x509::Certificate*>& certChain, CertInfo& topCertInfo, StaticString<cPasswordLen>& password)
+        = 0;
+
+    /**
+     * Removes certificate chain using top level certificate URL and password.
+     *
+     * @param certURL top level certificate URL.
+     * @param password certificate password.
```
al1img: owner password

include/aos/iam/certhandler.hpp
```diff
@@ -24,11 +195,81 @@ namespace certhandler {
...
-     * @return Error
+     * @param[out] certTypes result certificate types.
+     * @returns void.
+     */
+    void GetCertTypes(Array<StaticString<cCertificateTypeLen>>& certTypes);
+
+    /**
+     * Owns security storage.
+     *
+     * @param certTypes certificate type.
```
al1img, line 110: `certType`

include/aos/iam/certhandler.hpp
```diff
@@ -24,11 +195,81 @@ namespace certhandler {
...
+     * @param serial serial number.
+     * @returns RetWithError<CertInfo>.
+     */
+    RetWithError<CertInfo> GetCertificate(const String& certType, const Array<uint8_t>& issuer, const String& serial);
+
+    /**
+     * Creates a self signed certificate.
+     *
+     * @param certType certificate type.
+     * @param password certificate password.
```
al1img, line 158: owner password

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Certificate type name length.
+ */
+constexpr auto cCertificateTypeLen = AOS_IAM_CERTIFICATE_TYPE_NAME_LEN;
+
+/**
+ * Public/private keys generating algorithm
+ */
+enum class KeyGenAlgorithm { eRSA };
```
al1img, line 36: Use Aos::Enum

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+ */
+struct ModuleConfig {
+    /**
+     * Key generating algorithm.
+     */
+    KeyGenAlgorithm mKeyGenAlgorithm;
+    /**
+     * Maximum number of certificates for module.
+     */
+    unsigned mMaxCertificates = 0U;
```
al1img, line 97: Use `size_t` for such a number.
Leave it uninitialized.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+     */
+    StaticArray<ExtendedKeyUsage, cModuleKeyUsagesMaxCount> mExtendedKeyUsage;
+    /**
+     * Alternative DNS names.
+     */
+    StaticArray<StaticString<crypto::cDnsNameLen>, crypto::cAltDnsNamesCount> mAlternativeNames;
+    /**
+     * Skip certificate chain validation.
+     */
+    bool mSkipValidation = false;
```
al1img, line 109: Leave it uninitialized.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * Certificate type name length.
+ */
+constexpr auto cCertificateTypeLen = AOS_CONFIG_CERTMODULE_CERT_TYPE_NAME_LEN;
+
+/**
+ * Max number of invalid certificates per module.
+ */
+constexpr auto cInvalidCertsPerModule = AOS_CONFIG_CERTHANDLER_INVALID_CERT_PER_MODULE;
```
al1img, line 40: One const for all these items: invalid certs, invalid keys,  crtys per module should be enough. It should be the same const. And they should be moved to private CertModule modifier.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+     * @param certInfo info to compare.
+     * @return bool.
+     */
+    bool operator!=(const CertInfo& certInfo) const { return !operator==(certInfo); }
+};
+
+/**
+ * Platform dependent secure certificate storage.
+ */
+class SecureStorageItf {
```
al1img, line 155: Probably better to name it as `HSMItf`. HSM is a commonly used acronym: https://en.wikipedia.org/wiki/Hardware_security_module

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * Applies certificate chain to a module.
+     *
+     * @param certChain certificate chain.
+     * @param[out] topCertInfo info about a top certificate in a chain.
+     * @param[out] password owner password.
+     * @return Error.
+     */
+    virtual Error ApplyCertChain(
```
al1img, line 188: Rename to `ApplyCert`. Also I guess we can manage keeping password in CertModule instance. No need to return it here. In this case this method should look like:
```c++
virtual RetWithError<CertInfo> ApplyCertChain(const Array<crypto::x509::Certificate>& certChain)
```

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+    };
+
+    /**
+     * Creates a new object instance.
+     *
+     * @param csrProvider a reference to x509 csr provider.
+     * @param certProvider a reference to x509 certificate provider.
+     * @param secureStorage a reference to secure storage.
+     */
+    CertModule(crypto::x509::CSRProviderItf& csrProvider, crypto::x509::CertificateProviderItf& certProvider,
```
al1img, line 270: All interfaces should passed in `init` method.  Also, all commits should be buildable and should not break the current build and functionality: https://kb.epam.com/display/EPMDAEPRA/Coding+style+and+commits+submission

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * Module key usage name length.
+ */
+constexpr auto cModuleKeyUsageNameLen = AOS_CONFIG_CERTMODULE_KEY_USAGE_LEN;
+
+/**
+ * Maximum number of module key usages.
+ */
+constexpr auto cModuleKeyUsagesMaxCount = AOS_CONFIG_CERTMODULE_KEY_USAGE_MAX_COUNT;
```
al1img, line 30: Can it be moved to private ModuleConfig section?

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,380 @@
...
+#include "aos/iam/config.hpp"
+
+namespace aos {
+namespace iam {
+namespace certhandler {
+
+/**
+ * Module key usage name length.
+ */
+constexpr auto cModuleKeyUsageNameLen = AOS_CONFIG_CERTMODULE_KEY_USAGE_LEN;
```
al1img, line 25: Is it used?

include/aos/iam/config.hpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+#ifndef AOS_CONFIG_CERTMODULE_CERT_TYPE_NAME_LEN
+#define AOS_CONFIG_CERTMODULE_CERT_TYPE_NAME_LEN 16
+#endif
+
+/**
+ * Max number of key usages for a module.
+ */
+#ifndef AOS_CONFIG_CERTMODULE_KEY_USAGE_MAX_COUNT
+#define AOS_CONFIG_CERTMODULE_KEY_USAGE_MAX_COUNT 10
```
al1img, line 22: Currently it can be 2.

include/aos/iam/config.hpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT
+#define AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT 10
+#endif
+
+/**
+ * Max expected number of certificates in a chain stored in PEM file.
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_CERTS_CHAIN_SIZE
+#define AOS_CONFIG_CERTHANDLER_CERTS_CHAIN_SIZE 5
```
al1img, line 43: 3 by default

include/aos/iam/config.hpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_MAX_CERTS_PER_MODULE
+#define AOS_CONFIG_CERTHANDLER_MAX_CERTS_PER_MODULE 5
+#endif
+
+/**
+ * Max size of self signed certificate in PEM format (in bytes).
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_PEM_SELFSIGNED_CERT_SIZE
+#define AOS_CONFIG_CERTHANDLER_PEM_SELFSIGNED_CERT_SIZE 2048 + 1024
```
al1img, line 57: 4096?

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,266 @@
...
+public:
+    /**
+     * Returns public part of a private key.
+     */
+    const Array<uint8_t>& GetPublic() const { return mPublicKey; }
+
+    /**
+     * Returns private part of a private key.
+     */
+    const Array<uint8_t>& GetPrivate() const { return mPublicKey; }
```
al1img, line 115: This definitely not needed. Private key is not accessible in either way.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,266 @@
...
+struct Extension {
+    ObjectIdentifier                        mId;
+    StaticArray<uint8_t, cCertExtValueSize> mValue;
+};
+
+/**
+ * ASN1 Encodes string into ASN1 value
+ */
+
+inline void Encode(const String& /*src*/, Array<uint8_t>& /*asn1Value*/)
```
al1img, line 140: Could you put var names?

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,266 @@
...
+        = 0;
+
+    /**
+     * Creates certificates from a PEM blob.
+     *
+     * @param pemBlob raw certificate in a PEM format.
+     * @param[out] resultCerts result certificates.
+     * @result Error.
+     */
+    virtual Error CreateCertificatesFromPEM(const Array<uint8_t>& pemBlob, Array<Certificate>& resultCerts) = 0;
```
al1img, line 211: It actually, doesn't create it converts. So, `PEMToX509Cert` and `X509CertToPEM`  are good names here.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * Contains extra extensions applied to CSR.
+     */
+    StaticArray<asn1::Extension, cCertificateExtraExtCount> mExtraExtensions;
+};
+
+/**
+ * Provides interface to manage certificate requests.
+ */
+class CSRProviderItf {
```
al1img, line 244: No need different interface for each method. One `x509ProviderItf` would be enough.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,11 +16,209 @@ namespace certhandler {
...
-    LOG_DBG() << "Create key";
+}
+
+Error CertHandler::RegisterModule(CertModule& module)
+{
+    LockGuard lock(mMutex);
+    LOG_INF() << "Register module with cert type: " << module.GetCertType();
+
+    Error error = module.Init();
+    if (error != Error::Enum::eNone) {
```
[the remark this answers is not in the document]
mykola-kobets-epam: Fixed

include/aos/common/crypto.hpp
al1img: You can supress it by adding `(void)src;' in function implementation.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,226 @@
...
+public:
+    /**
+     * Returns public part of a private key.
+     */
+    const Array<uint8_t>& GetPublic() const { return mPublicKey; }
+
+    /**
+     * Returns private part of a private key.
+     */
+    const Array<uint8_t>& GetPrivate() const { return mPrivateKey; }
```
al1img, line 93: Again, private key should not be exposed as public API in either case. So remove this method as well as cCertPrivKeyLen.

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+ * Certificate serial number len.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_SERIAL_LEN
+#define AOS_CONFIG_CRYPTO_CERT_SERIAL_LEN 40
+#endif
+
+/**
+ * Certificate issuer len.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_ISSUER_ID_LEN
```
al1img, line 182: It is not ID. Issuer is rather a struct with additional fields inside. Rename to `AOS_CONFIG_CRYPTO_CERT_ISSUER_SIZE`

40 definitely is not enough: let's put 256 for now.

Use `LEN` for string values and `SIZE` for other.

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_PRIV_KEY_LEN
+#define AOS_CONFIG_CRYPTO_CERT_PRIV_KEY_LEN 2048
+#endif
+
+/**
+ * Certificate serial number len.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_SERIAL_LEN
+#define AOS_CONFIG_CRYPTO_CERT_SERIAL_LEN 40
```
al1img, line 176: Looks like it should be 20 byte len: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.2

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
 #define AOS_CONFIG_MONITORING_POLL_PERIOD_SEC 10
 #endif
 
+/**
+ * Certificate public key len.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_PUB_KEY_LEN
```
al1img, line 161: Remove period at the end of commit message.

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+ */
+#ifndef AOS_CONFIG_CRYPTO_DNS_NAME_LEN
+#define AOS_CONFIG_CRYPTO_DNS_NAME_LEN 42
+#endif
+
+/**
+ * Max number of alternative DNS names for a module.
+ */
+#ifndef AOS_CONFIG_CRYPTO_ALT_DNS_NAMES_MAX_COUNT
+#define AOS_CONFIG_CRYPTO_ALT_DNS_NAMES_MAX_COUNT 10
```
al1img, line 197: we don't need 10. 4 is max for now.

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+#endif
+
+/**
+ * Certificate subject size.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_SUBJECT_SIZE
+#define AOS_CONFIG_CRYPTO_CERT_SUBJECT_SIZE 24
```
al1img, line 204: Don't think 24 is enough. 256 for now. Also, it probably should be equal to `AOS_CONFIG_CRYPTO_CERT_ISSUER_SIZE`. Need to be clarified.

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+ */
+#ifndef AOS_CONFIG_CRYPTO_EXTRA_EXTENSIONS_COUNT
+#define AOS_CONFIG_CRYPTO_EXTRA_EXTENSIONS_COUNT 10
+#endif
+
+/**
+ * Maximum length of numeric string representing ASN.1 Object Identifier.
+ */
+#ifndef AOS_CONFIG_CRYPTO_OBJECT_ID_LEN
+#define AOS_CONFIG_CRYPTO_OBJECT_ID_LEN 16
```
al1img, line 225: Add asn1 to name, something like `AOS_CONFIG_CRYPTO_ASN1_OBJECT_ID_LEN`

include/aos/common/config.hpp
```diff
@@ -155,4 +155,88 @@
...
+#endif
+
+/**
+ * Maximum size(in bytes) of a certificate extension value.
+ */
+#ifndef AOS_CONFIG_CRYPTO_CERT_EXTENSION_VALUE_SIZE
+#define AOS_CONFIG_CRYPTO_CERT_EXTENSION_VALUE_SIZE 16
```
al1img, line 232: Same: `AOS_CONFIG_CRYPTO_ASN1_VALUE_MAX_SIZE`

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,378 @@
...
+        = 0;
+
+    /**
+     * Removes certificate chain using top level certificate URL and password.
+     *
+     * @param certURL top level certificate URL.
+     * @param password owner password.
+     * @return Error.
+     */
+    virtual Error RemoveCertChain(const String& certURL, const String& password) = 0;
```
al1img, line 201: Rename `RemoveCertChain` to `RemoveCert` it should match `ApplyCert`.

include/aos/common/time.hpp
```diff
@@ -0,0 +1,44 @@
...
+    uint64_t UnixNano() const { return 0ULL; }
+
+    bool operator<(const Time& /*right*/) const { return true; }
+    bool operator==(const Time& /*right*/) const { return true; }
+    bool operator!=(const Time& time) const { return !operator==(time); }
+
+    static Time Now() { return Time(); }
+};
+
+template <class Stream>
```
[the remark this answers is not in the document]
al1img: friend function is ok in this case.

include/aos/common/time.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace time {
+
+/**
+ * A certain point in time.
+ * TODO: Add implementation
+ */
+struct Time {
+    bool IsZero() const { return true; }
+
+    Time AddYears(int /*year*/) { return Time(); }
```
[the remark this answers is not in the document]
al1img: Use (void)varName inside function.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,378 @@
...
+};
+
+/**
+ * General certificate information.
+ */
+struct CertInfo {
+    /**
+     * Certificate issuer name.
+     */
+    StaticString<crypto::cCertIssuerLen> mIssuer;
```
al1img, line 104: We should not use string for issuer and serial number. We should use StaticArray<uint8_t> instead.

include/aos/common/base64.hpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (C) 2023 Renesas Electronics Corporation.
+ * Copyright (C) 2023 EPAM Systems, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef AOS_COMMON_BASE64_HPP_
```
al1img, line 8: We don't need base64 encoding. Let's discuss offline.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+    mStorage = &storage;
+}
+
+Error CertHandler::RegisterModule(CertModule& module)
+{
+    LockGuard lock(mMutex);
+
+    LOG_INF() << "Register module with cert type: " << module.GetCertType();
+
+    // TODO: clarify. If module takes external interfaces in Init method, so it ahs to be initialized outside. right?
```
al1img, line 35: Right, https://github.com/aoscloud/aos_core_zephyr/blob/main/src/main.cpp `app.Init()`.
https://github.com/aoscloud/aos_core_zephyr/blob/main/src/app/app.cpp

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,11 +16,209 @@ namespace certhandler {
...
+    for (const auto module : mModules) {
+        certTypes.PushBack(module->GetCertType());
+    }
+}
+
+Error CertHandler::SetOwner(const String& certType, const String& password)
+{
+    LockGuard lock(mMutex);
+
+    auto* module = FindModule(certType);
```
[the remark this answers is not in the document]
al1img: Sure, you can. Minor comment: whay do you need to return just pointer and then check if it is null if `Array::Find` already return this error. So, `FindModule` may have the following format:  `aos::RetWithError<CertModule*> FindModule(const String& certType)`.

```c++
auto module = FindModule(certType);
if (!module.mError.IsNone()) {
    return module.mError;
}
```

Could this method be const also?

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+
+RetWithError<CertInfo> CertHandler::GetCertificate(
+    const String& certType, const Array<uint8_t>& issuer, const String& serial)
+{
+    LockGuard lock(mMutex);
+
+    LOG_DBG() << "Get certificate of a certType: " << certType << " with serial: " << serial;
+
+    if (serial == "") {
+        StaticArray<CertInfo, cCertsPerModule> certsInfo;
```
al1img, line 133: Add empty line after var definition.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+        const auto                             err = mStorage->GetCertsInfo(certType, certsInfo);
+        if (!err.IsNone()) {
+            return {CertInfo {}, err};
+        }
+
+        if (certsInfo.Size() == 0) {
+            return {CertInfo {}, Error::Enum::eNotFound};
+        }
+
+        CertInfo result = {};
```
al1img, line 143: ditto

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+}
+
+void CertHandler::OnCertRemoved(const String& certType, const String& certURL)
+{
+    mStorage->RemoveCertInfo(certType, certURL);
+}
+
+CertModule* CertHandler::FindModule(const String& certType)
+{
+    auto module = mModules.Find([certType](CertModule* module) { return module->GetCertType() == certType; });
```
al1img, line 194: Empty line before return.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+    return !module.mError.IsNone() ? nullptr : *module.mValue;
+}
+
+Error CertHandler::SyncValidCerts(CertModule& module)
+{
+    const Array<CertInfo>& validCerts = module.GetValidCerts();
+
+    StaticArray<CertInfo, cCertsPerModule> certsInStorage;
```
al1img, line 202: Empty line after var definition.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +17,231 @@ namespace certhandler {
...
+    auto                                   err = mStorage->GetCertsInfo(module.GetCertType(), certsInStorage);
+    if (!err.IsNone() || err != Error::Enum::eNotFound) {
+        return err;
+    }
+
+    // Add module certificates to storage.
+    for (const auto& moduleCert : validCerts) {
+
+        CertInfo* storedCert = nullptr;
```
al1img, line 211: ditto

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,291 @@
...
+    return mSecureStorage.CreateKey(password, mModuleConfig.mKeyGenAlgorithm);
+}
+
+Error CertModule::CreateCSR(const String& subject, const crypto::PrivateKey& privKey, crypto::x509::CSR& csr)
+{
+    crypto::x509::CSR templ;
+    templ.mSubject = subject;
+    templ.mDnsNames = mModuleConfig.mAlternativeNames;
+
+    static constexpr auto oidExtensionExtendedKeyUsage = "2.5.29.37";
```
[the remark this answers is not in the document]
al1img: I'm just asking if we should move this const to crypto.hpp

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+            LOG_WRN() << "Unexpected extended key usage value: " << extKeyUsage.ToString();
+            break;
+        }
+    }
+
+    if (!oid.IsEmpty()) {
+        StaticArray<uint8_t, crypto::cCertExtValueSize> extValue;
+        crypto::asn1::Encode(oid, extValue);
+
+        crypto::asn1::Extension ext;
```
al1img, line 109: Reformat as following:
```c++
StaticArray<uint8_t, crypto::cCertExtValueSize> extValue;
crypto::asn1::Extension ext;

ext.mId = oidExtensionExtendedKeyUsage;
crypto::asn1::Encode(oid, extValue);
ext.mValue.PushBack(extValue);

templ.mExtraExtensions.PushBack(ext);
```

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+        templ.mExtraExtensions.PushBack(ext);
+    }
+
+    return mX509Provider->CreateCSR(templ, privKey, csr);
+}
+
+Error CertModule::ApplyCert(const Array<uint8_t>& pemCert, CertInfo& info)
+{
+    StaticArray<crypto::x509::Certificate, cCertChainSize> certificates;
+    auto                                                   err = mX509Provider->PEMToX509Certs(pemCert, certificates);
```
al1img, line 122: Empty line after var definition.

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    err = CheckCertChain(certificates);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    StaticString<cPasswordLen> password;
```
al1img, line 132: ditto

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+}
+
+Error CertModule::CreateSelfSignedCert(const String& password)
+{
+    auto key = CreateKey(password);
+    if (!key.mError.IsNone()) {
+        return key.mError;
+    }
+
+    static constexpr auto cValidSelfSignedCertPeriod = 100; // in years
```
al1img, line 152: Move to private block of cert module. I prefer all consts in one place. As we introduce `Time` module. It should have unit constant inside: such a Time::Hour, Time::Minute, etc. So, value of this const should lool like:
cValidSelfSignedCertPeriod = 100 * 365 * 24 * Time::Hour. We can't have const like Time::Year unfortunately.

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+    const uint64_t serial = time::Time::Now().UnixNano();
+
+    crypto::x509::Certificate templ;
```
al1img, line 156: Empty line after var definition.

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+    templ.mSerial = Array<uint8_t>(reinterpret_cast<const uint8_t*>(serial), sizeof(serial));
+    templ.mNotBefore = time::Time::Now();
+    templ.mNotAfter = time::Time::Now().AddYears(cValidSelfSignedCertPeriod);
+    templ.mSubject = "Aos Core";
+
+    DynamicArray<uint8_t, cMaxSelfSignedCertSize> pemCert;
```
al1img, line 162: ditto

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+{
+    StaticArray<CertInfo, cCertsPerModule> certsInfo = mValidCerts;
+
+    if (certsInfo.Size() > mModuleConfig.mMaxCertificates) {
+        LOG_WRN() << "Current cert count exceeds max count: " << certsInfo.Size() << " > "
+                  << mModuleConfig.mMaxCertificates << ". Remove old certificates";
+    }
+
+    while (certsInfo.Size() > mModuleConfig.mMaxCertificates) {
+        time::Time minTime;
```
al1img, line 216: ditto

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,287 @@
...
+        return Error::Enum::eNotFound;
+    }
+
+    for (const auto& cert : chain) {
+        LOG_DBG() << "Check certificate chain: "
+                  << "[ issuer: " << cert.mIssuer << ", subject: " << cert.mSubject << " ]";
+    }
+
+    // TODO: clairfy whther Issuer/Subject test is ok (source impl. tested RawIssuer/RawSubject instead).
+    size_t currentCert = 0;
```
al1img, line 257: ditto

src/iam/certhandler/certhandler.cpp
mykola-kobets-epam: fixed

src/iam/certhandler/certhandler.cpp
mykola-kobets-epam: fixed

src/iam/certhandler/certmodule.cpp
mykola-kobets-epam: fixed

src/iam/certhandler/certmodule.cpp
mykola-kobets-epam: fixed

include/aos/iam/certmodule.hpp
mykola-kobets-epam: Done

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,386 @@
...
+    /**
+     * Creates a self signed certificate.
+     *
+     * @param password owner password.
+     * @returns Error.
+     */
+    Error CreateSelfSignedCert(const String& password);
+
+private:
+    static constexpr auto cCertChainSize = AOS_CONFIG_CERTHANDLER_CERTS_CHAIN_SIZE;
```
al1img, line 355: Could you move all private members to implementation commit? Actually, all private members should be defined in the commit they are used. Otherwise, it is hard to review.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,253 @@
...
+    virtual void CreateDN(const String& commonName, Array<uint8_t>& result) = 0;
+
+    /**
+     * Returns text representation of x509 distinguished name(DN).
+     *
+     * @param dn source binary representation of DN.
+     * @param[out] result DN text representation.
+     * @result void.
+     */
+    virtual void DN2String(const Array<uint8_t>& dn, String& result) = 0;
```
al1img, line 241: `DNToString` or rename `PEMToX509Certs` accordingly. `DNToString` is more preferable. We don't use `2` as `To` usually in this project.

It should return err.

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +16,227 @@ namespace certhandler {
...
+    StaticArray<CertInfo, cCertsPerModule> certsInStorage;
+
+    auto err = mStorage->GetCertsInfo(module.GetCertType(), certsInStorage);
+    if (!err.IsNone() && err != Error::Enum::eNotFound) {
+        return err;
+    }
+
+    // Add module certificates to storage.
+    for (const auto& moduleCert : validCerts) {
+
```
al1img, line 203: Remove empty line.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,253 @@
...
+    virtual Error CreateCSR(const CSR& templ, const PrivateKey& privKey, CSR& resultCSR) = 0;
+
+    /**
+     * Constructs x509 distinguished name(DN) from the argument list.
+     *
+     * @param comName common name part of a result DN.
+     * @param[out] result result DN.
+     * @result Error.
+     */
+    virtual void CreateDN(const String& commonName, Array<uint8_t>& result) = 0;
```
al1img, line 232: It should return err.

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,295 @@
...
+        auto err = mHSM->RemoveCert(info->mCertURL, password);
+        if (!err.IsNone()) {
+            return err;
+        }
+
+        err = mHSM->RemoveKey(info->mKeyURL, password);
+        if (!err.IsNone()) {
+            return err;
+        }
+        if (mCertListener != nullptr) {
```
al1img, line 239: Add empty line

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,295 @@
...
+            mCertListener->OnCertRemoved(GetCertType(), info->mCertURL);
+        }
+        mValidCerts.Remove(info);
```
al1img, line 242: Ditto

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+const String& CertModule::GetCertType() const
+{
+    return mCertType;
+}
+
+const Array<CertInfo>& CertModule::GetValidCerts() const
+{
+    return mValidCerts;
```
al1img, line 51: We didn't keep it in memory just to reduce memory usage. We have a bunch of certificates probably up 15-20 for each node. It could be an option as well but in this case, it is useless to keep these certificated in storage at all. If we keep this info in memory, no need to read it each time from DB (disk).  `CertModule` should provide required info for `CertHandler`.

include/aos/iam/certmodule.hpp
```diff
@@ -0,0 +1,386 @@
...
+
+/**
+ * Certificate type name length.
+ */
+constexpr auto cCertTypeLen = AOS_CONFIG_CERTMODULE_CERT_TYPE_NAME_LEN;
+
+/**
+ * Max number of IAM certificates per module.
+ */
+constexpr auto cCertsPerModule = AOS_CONFIG_CERTHANDLER_MAX_CERTS_PER_MODULE;
```
al1img, line 35: It looks like this constant should be the same as `mMaxCertificates+1` in the module config. Probably we should add template parameter:

```c++
template <size_t cMaxCerts>
CertModule(...)
```

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +16,227 @@ namespace certhandler {
...
+}
+
+Error CertHandler::RegisterModule(CertModule& module)
+{
+    LockGuard lock(mMutex);
+
+    LOG_INF() << "Register module with cert type: " << module.GetCertType();
+
+    module.SetListener(*this);
+    mModules.PushBack(&module);
```
al1img, line 35: Check error on all array push

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +16,227 @@ namespace certhandler {
...
+}
+
+void CertHandler::GetCertTypes(Array<StaticString<cCertTypeLen>>& certTypes)
+{
+    LockGuard lock(mMutex);
+
+    LOG_DBG() << "Get all registered IAM certificate types";
+
+    for (const auto module : mModules) {
+        certTypes.PushBack(module->GetCertType());
```
al1img, line 47: Check error.  Also this function should return error I guess.

include/aos/iam/certmodule.hpp
mykola-kobets-epam: Basically that's difficult because of two reasons:  git gui cant unstage lines for a new file, an alternative command line approach is a way more difficult to use, also it would involve some modification to dependent config values.

al1img: Actually, it is quite easy to achieve using interactive rebase: cut it in this commit and insert in another one. Do not understand about config values. You can keep config values in this commit.

src/iam/certhandler/certmodule.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+const String& CertModule::GetCertType() const
+{
+    return mCertType;
+}
+
+RetWithError<CertInfo> CertModule::GetCertificate(const Array<uint8_t>& issuer, const Array<uint8_t>& serial)
+{
+    if (serial.IsEmpty()) {
+        const auto err = mStorage->GetCertsInfo(GetCertType(), mTmpCertsInStorage);
```
al1img, line 48: Let's use just `auto` for `err`. It could be reused normally.

tests/iam/CMakeLists.txt
```diff
@@ -11,7 +11,9 @@ set(TARGET aosiamcpp_test)
 # Sources
 # ######################################################################################################################
 
-set(SOURCES ${CMAKE_SOURCE_DIR}/src/iam/certhandler/certhandler.cpp certhandler/certhandler_test.cpp)
+set(SOURCES ${CMAKE_SOURCE_DIR}/src/iam/certhandler/certhandler.cpp
```
al1img: Not relevant change.

tests/iam/certhandler/mocks/hsm_itf.hpp
```diff
@@ -0,0 +1,48 @@
+/*
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

tests/iam/certhandler/mocks/storage_itf.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (C) 2023 Renesas Electronics Corporation.
+ * Copyright (C) 2023 EPAM Systems, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef AOS_MOCK_STORAGE_ITF_HPP_
+#define AOS_MOCK_STORAGE_ITF_HPP_
+
+#include "aos/iam/certhandler.hpp"
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

tests/iam/certhandler/mocks/x509provider_itf.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (C) 2023 Renesas Electronics Corporation.
+ * Copyright (C) 2023 EPAM Systems, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef AOS_MOCK_X509PROVIDER_ITF_HPP_
+#define AOS_MOCK_X509PROVIDER_ITF_HPP_
+
+#include "aos/common/crypto.hpp"
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

src/iam/certhandler/certhandler.cpp
```diff
@@ -16,6 +16,141 @@ namespace certhandler {
...
+
+    LOG_DBG() << "Get all registered IAM certificate types";
+
+    for (const auto module : mModules) {
+        Error err = certTypes.PushBack(module->GetCertType());
+        if (!err.IsNone()) {
+            return AOS_ERROR_WRAP(err.Value());
+        }
+    }
+    return Error::Enum::eNone;
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

src/iam/certhandler/certmodule.cpp
mykola-kobets-epam: fixed.

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+
+    static const String cPKCS11Type;
+    static const String cSWType;
+    static const String cTPMType;
+
+    static const String cPassword;
+
+    static const Array<uint8_t> cIssuer;
+    static const Array<uint8_t> cSubject;
+    static crypto::PrivateKey   cPrivateKey;
```
al1img, line 65: Move const to top of protected sections: first const, methods, vars

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+
+        StaticArray<uint8_t, 1> pemCert;
+        CertInfo                certInfo;
+        EXPECT_EQ(Error::Enum::eNone, mCertHandler.ApplyCertificate(cPKCS11Type, pemCert, certInfo));
+    }
+
+    // certmodule class dependendcies
+    MockHSMItf                    mPKCS11, mSW, mTPM;
+    crypto::x509::MockProviderItf mX509Provider;
+    // cerhandler class dependencies
```
al1img, line 50: storage now certModule dep

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+
 using aos::Error;
-using aos::iam::certhandler::CertHandler;
+using namespace aos;
+using namespace aos::iam::certhandler;
+using namespace ::testing;
+
+class CertHandlerTest : public ::testing::Test {
+protected:
+    virtual void SetUp()
```
al1img, line 25: Is it override method? If os, add override keyword.

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+    {
+        EXPECT_CALL(mX509Provider, DNToString(_, _))
+            .WillRepeatedly(Invoke([](const Array<uint8_t>& dn, String& result) {
+                result = reinterpret_cast<const char*>(dn.begin());
+                return Error::Enum::eNone;
+            }));
+    }
+
+    void ApplyCert(const Array<crypto::x509::Certificate>& certChain, const CertInfo& devCertInfo)
```
al1img, line 34: Why it is moved to separate method?

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+
+    EXPECT_CALL(mPKCS11, ValidateCertificates(_, _, _)).WillOnce(Return(Error::Enum::eNone));
+    ASSERT_EQ(Error::Enum::eNone, pkcs11.Init(mX509Provider, mPKCS11, mStorage));
+    ASSERT_EQ(Error::Enum::eNone, mCertHandler.RegisterModule(pkcs11));
+
+    EXPECT_CALL(mPKCS11, SetOwner(cPassword)).WillOnce(Return(Error::Enum::eNone));
+    ASSERT_EQ(Error::Enum::eNone, mCertHandler.SetOwner(cPKCS11Type, cPassword));
+}
+
+TEST_F(CertHandlerTest, SetOwner_ModuleNotFound)
```
al1img, line 128: Do not use _ in names

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+    EXPECT_CALL(mPKCS11, ValidateCertificates(_, _, _)).WillOnce(Return(Error::Enum::eNone));
+    ASSERT_EQ(Error::Enum::eNone, pkcs11.Init(mX509Provider, mPKCS11, mStorage));
+    ASSERT_EQ(Error::Enum::eNone, mCertHandler.RegisterModule(pkcs11));
+
+    EXPECT_CALL(mStorage, RemoveAllCertsInfo(cPKCS11Type)).WillOnce(Return(Error::Enum::eNone));
+    EXPECT_CALL(mPKCS11, Clear()).WillOnce(Return(Error::Enum::eNone));
+    ASSERT_EQ(Error::Enum::eNone, mCertHandler.Clear(cPKCS11Type));
+}
+
+TEST_F(CertHandlerTest, Clear_ModuleNotFound)
```
al1img, line 157: ditto

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+    using crypto::x509::CSR;
+
+    Extension                 extension = {"2.5.29.37", {}};
+    StaticArray<Extension, 1> extensions = Array<Extension>(&extension, 1);
+
+    auto templ = AllOf(Field(&CSR::mSubject, cSubject), Field(&CSR::mDnsNames, moduleConfig.mAlternativeNames),
+        Field(&CSR::mExtraExtensions, extensions));
+    EXPECT_CALL(mX509Provider, CreateCSR(templ, _, _)).WillOnce(Return(Error::Enum::eNone));
+
+    CSR csr;
```
al1img, line 262: Empty line after var definition.

tests/iam/certhandler/mocks/hsm.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace iam {
+namespace certhandler {
+
+/**
+ * StorageItf provides API to store/retrieve certificates info.
+ */
+class MockHSMItf : public HSMItf {
+public:
+    MOCK_METHOD(Error, SetOwner, (const String& password), (override));
+
```
al1img, line 24: Remove empty lines between methods.

tests/iam/certhandler/mocks/storage.hpp
```diff
@@ -0,0 +1,39 @@
...
+namespace iam {
+namespace certhandler {
+
+/**
+ * StorageItf provides API to store/retrieve certificates info.
+ */
+class MockStorageItf : public StorageItf {
+public:
+    MOCK_METHOD(Error, AddCertInfo, (const String& certType, const CertInfo& certInfo), (override));
+
```
al1img, line 24: ditto

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+    {
+        EXPECT_CALL(mX509Provider, PEMToX509Certs(_, _))
+            .WillOnce(DoAll(SetArgReferee<1>(certChain), Return(Error::Enum::eNone)));
+        EXPECT_CALL(mPKCS11, ApplyCert(_, _, _))
+            .WillOnce(DoAll(SetArgReferee<1>(devCertInfo), SetArgReferee<2>(cPassword), Return(Error::Enum::eNone)));
+        EXPECT_CALL(mStorage, AddCertInfo(cPKCS11Type, devCertInfo)).WillOnce(Return(Error::Enum::eNone));
+
+        StaticArray<uint8_t, 1> pemCert;
+        CertInfo                certInfo;
+        EXPECT_EQ(Error::Enum::eNone, mCertHandler.ApplyCertificate(cPKCS11Type, pemCert, certInfo));
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

tests/iam/certhandler/certhandler_test.cpp
mykola-kobets-epam: fixed

tests/iam/certhandler/certhandler_test.cpp
mykola-kobets-epam: done

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,523 @@
...
+const Array<uint8_t> CertHandlerTest::cIssuer
+    = StringToDN("C = UA, OU = My Digest Company, CN = Developer Relations Cert");
+const Array<uint8_t> CertHandlerTest::cSubject = StringToDN("UID = XMM9NE5AEO, OU = Worker, C = UA");
+crypto::PrivateKey   CertHandlerTest::cPrivateKey;
+
+TEST_F(CertHandlerTest, GetCertTypes)
+{
+    EXPECT_CALL(mStorage, GetCertsInfo(_, _)).WillRepeatedly(Return(Error::Enum::eNone));
+
+    CertModule pkcs11 = {cPKCS11Type, cDefaultConfig};
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

tests/iam/CMakeLists.txt
mykola-kobets-epam: otherwise build fails because of UT, in the following commit it is replaced with link_library to iam library.

include/aos/common/crypto.hpp
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Max number of alternative names for a module.
+ */
+constexpr auto cAltDnsNamesCount = AOS_CONFIG_CRYPTO_ALT_DNS_NAMES_MAX_COUNT;
+
+/**
+ * Certificate subject size.
+ */
+constexpr auto cCertSubjSize = AOS_CONFIG_CRYPTO_CERT_ISSUER_SIZE;
+
```
mykola-kobets-epam, line 49: Now AOS_CONFIG_CRYPTO_CERT_ISSUER_SIZE serves as subject size as well

src/iam/certhandler/modules/certmodule.cpp
```diff
@@ -0,0 +1,418 @@
...
+    if (mModuleConfig.mSkipValidation) {
+        LOG_WRN() << "Skip validation. type=" << GetCertType();
+
+        return Error::Enum::eNone;
+    }
+
+    auto validCerts = MakeUnique<ModuleCertificates>(&mAllocator);
+    if (!validCerts) {
+        return AOS_ERROR_WRAP(Error::Enum::eNoMemory);
+    }
```
mykola-kobets-epam, line 41: Currently it's not clear about memory allocation error handling: return an error / use assert. Current impl follows an approach with PushBack, ie - return an error.

src/iam/certhandler/modules/certmodule.cpp
```diff
@@ -0,0 +1,418 @@
...
+{
+    auto templ = MakeUnique<crypto::x509::CSR>(&mAllocator);
+    if (!templ) {
+        AOS_ERROR_WRAP(Error::Enum::eNone);
+    }
+
+    templ->mSubject = subject;
+    templ->mDnsNames = mModuleConfig.mAlternativeNames;
+
+    StaticArray<crypto::asn1::ObjectIdentifier, crypto::cCertExtraExtCount> oids;
```
mykola-kobets-epam, line 126: Small objects are still located on the stack, large - in allocator.

include/aos/iam/config.hpp
```diff
@@ -0,0 +1,60 @@
...
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_MAX_CERTS_PER_MODULE
+#define AOS_CONFIG_CERTHANDLER_MAX_CERTS_PER_MODULE 5
+#endif
+
+/**
+ * Max size of self signed certificate in PEM format (in bytes).
+ */
+#ifndef AOS_CONFIG_CERTHANDLER_PEM_SELFSIGNED_CERT_SIZE
+#define AOS_CONFIG_CERTHANDLER_PEM_SELFSIGNED_CERT_SIZE 4096
```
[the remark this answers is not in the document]
al1img: Yes, in this particular case, but in general the size should be the same.

src/iam/certhandler/modules/certmodule.cpp
al1img: The idea is: with static allocator you have preallocated memory and it is almost used by this particular class that is fully covered with unit tests. And if your allocator memory is not enough it will be identified and fixed with unit testing. That's why it is not mandatory to check out of memory error. Push to array is performed on external for current class/module array and not covered with unit testing. So, in this case checking memory error is mandatory.

include/aos/iam/modules/certmodule.hpp
```diff
@@ -0,0 +1,420 @@
...
+    static constexpr auto cDNStringLen = AOS_CONFIG_CERTHANDLER_DN_STRING_LEN;
+    static constexpr auto cValidSelfSignedCertPeriod = time::Years(100);
+
+    static constexpr auto cOidExtensionExtendedKeyUsage = "2.5.29.37";
+    static constexpr auto cOidExtKeyUsageClientAuth = "1.3.6.1.5.5.7.3.1";
+    static constexpr auto cOidExtKeyUsageServerAuth = "1.3.6.1.5.5.7.3.2";
+
+    using StorageCerts = StaticArray<CertInfo, cCertsPerModule>;
+    using DNString = StaticString<cDNStringLen>;
+    using ModuleCertificates = StaticArray<CertInfo, cCertsPerModule>;
```
al1img, line 391: Why do we need different using for the same types? Name it to cover the both cases: `ModuleCerts` for example.

include/aos/iam/modules/certmodule.hpp
```diff
@@ -0,0 +1,420 @@
...
+    crypto::x509::ProviderItf* mX509Provider;
+    HSMItf*                    mHSM;
+    StorageItf*                mStorage;
+
+    StaticString<cCertTypeLen> mCertType;
+    ModuleConfig               mModuleConfig;
+
+    StaticArray<StaticString<cURLLen>, cCertsPerModule> mInvalidCerts, mInvalidKeys;
+
+    StaticAllocator<Max(sizeof(ModuleCertificates) + sizeof(StorageCerts), sizeof(crypto::x509::CSR),
```
al1img, line 410: CertModule::Init                    -> ModuleCertificates + StorageCerts
CertModule::GetCertificate -> StorageCerts
CertModule::CreateCSR       -> crypto::x509::CSR
CertModule::ApplyCert        -> CertificateChain + 2 * StorageCerts + 2 * DNString

src/iam/certhandler/modules/certmodule.cpp
```diff
@@ -0,0 +1,418 @@
...
+        return err;
+    }
+
+    err = mX509Provider->CreateDN("Aos Core", templ->mIssuer);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    auto pemCert = MakeUnique<SelfSignedCertificate>(&mAllocator);
+    auto certInfo = MakeUnique<CertInfo>(&mAllocator);
```
al1img, line 220: CertInfo is ok to keep in stack. Generally, your function should not exceed 2k of stack. It is rule for now.

src/iam/certhandler/modules/certmodule.cpp
```diff
@@ -0,0 +1,418 @@
...
+}
+
+Error CertModule::CheckCertChain(const Array<crypto::x509::Certificate>& chain)
+{
+    if (chain.IsEmpty()) {
+        return AOS_ERROR_WRAP(Error::Enum::eNotFound);
+    }
+
+    for (const auto& cert : chain) {
+        auto issuer = MakeUnique<DNString>(&mAllocator);
```
al1img, line 325: DNString is ok to keep in stack.

tests/iam/certhandler/certhandler_test.cpp
```diff
@@ -6,15 +6,541 @@
...
-//     CertHandler handler;
+using namespace aos;
+using namespace aos::iam::certhandler;
+using namespace testing;
 
-//     EXPECT_TRUE(handler.CreateKey().IsNone());
-// }
+/***********************************************************************************************************************
+ * Suite
+ ********************************************************************ma**************************************************/
```
al1img, line 24: Remove `ma`

src/iam/certhandler/modules/certmodule.cpp
mykola-kobets-epam: For now arrays with certificates and pem blobs will be stored in the allocator.

include/aos/iam/config.hpp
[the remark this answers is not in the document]
mykola-kobets-epam: ok, fixed

include/aos/iam/certhandler.hpp
```diff
@@ -24,11 +25,98 @@ namespace certhandler {
...
+
+    /**
+     * Destroys certificate handler object instance.
+     */
+    virtual ~CertHandler();
+
+private:
+    CertModule* FindModule(const String& certType) const;
+
+    static constexpr auto cIAMCertModulesMaxCount = AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT;
```
al1img, line 116: Group all memebers as following: consts, methods, variables.

include/aos/iam/modules/certmodule.hpp
```diff
@@ -0,0 +1,417 @@
...
+    /**
+     * Creates a self signed certificate.
+     *
+     * @param password owner password.
+     * @returns Error.
+     */
+    Error CreateSelfSignedCert(const String& password);
+
+private:
+    static constexpr auto cCertChainSize = AOS_CONFIG_CERTHANDLER_CERTS_CHAIN_SIZE;
```
al1img, line 379: cPEMCertSize is common crypto const. Make it public in common.

al1img (approved):

MykolaSolyanko (approved):

PR #47, opened 2023-11-08
[iam] pkcs11 cert module initial implementation

mykola-kobets-epam:


include/aos/common/config.hpp
```diff
@@ -232,4 +232,67 @@
 #define AOS_CONFIG_CRYPTO_CERT_KEY_ID_SIZE 24
 #endif
 
+/**
```
al1img, line 235: Remove period at the end of commit message.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,952 @@
...
+        return AOS_ERROR_WRAP(Error::Enum::eInvalidConfiguration);
+    }
+
+    err = pkcs11.InitLibrary(mConfig.mLibrary);
+    if (!err.IsNone()) {
+        return AOS_ERROR_WRAP(err);
+    }
+    mPKCS11 = &pkcs11;
+
+    SetTokenLabel();
```
al1img, line 54: Setting private member in this way decrease code readability IMHO.  To understand that  `SetTokenLabel` actually sets `mTokenLabel`, I have to go inside this function. `mTokenLabel = GetTokenLabel()` explicitly shows that at this place I set private member `mTokenLabel`.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,952 @@
...
+    } else {
+        LOG_DBG() << "No owned token found";
+    }
+
+    return err;
+}
+
+Error PKCS11::SetOwner(const String& password)
+{
+    Error err = SetSlotID();
```
al1img, line 82: auto err

include/aos/common/crypto.hpp
```diff
@@ -247,6 +262,717 @@ class ProviderItf {
...
+
+/**
+ * Converts a public key to PKIX, ASN.1 DER form. The encoded public key is a SubjectPublicKeyInfo structure (see RFC
+ * 5280, Section 4.1).
+ *
+ * @param pubKey public key.
+ * @param[out] pkixPubKey public key in PKIX, ASN.1 DER form.
+ * @return Error.
+ */
+inline Error MarshalPKIXPublicKey(const Array<uint8_t>& pubKey, Array<uint8_t>& pkixPubKey)
```
al1img, line 339: It is part of x509 namespace. Not sure we need it in c++ implementation.

include/aos/common/crypto.hpp
```diff
@@ -247,6 +262,717 @@ class ProviderItf {
...
+ */
+static constexpr auto cPINLength = AOS_CONFIG_CRYPTO_PKCS11_PIN_LEN;
+
+/**
+ * Generates random unique PIN.
+ *
+ * @param[out] pin result pin.
+ * @return Error.
+ */
+inline Error GenPIN(String& pin)
```
al1img, line 324: This function could be useful not only in pkcs11 context. We have to move it to som common part.

include/aos/common/uuid.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Returns a new SHA1 UUID based on the supplied name space and data.
+ *
+ * @param space name space.
+ * @param data data.
+ * @param[out] dst result UUID.
+ * @return Error.
+ */
+template <class T>
```
al1img, line 51: Why we need template here? I suppose it should be array of byte. RetWithError.

include/aos/common/uuid.hpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+using UUID = StaticArray<uint8_t, cUUIDLen>;
+
+/**
+ * Creates unique UUID.
+ *
+ * @param dst result uuid.
+ * @return Error.
+ */
+inline Error CreateUUID(UUID& dst)
```
al1img, line 37: We definitely could return it through stack: `RetWithError`.
Also, don't think all these functions will be inlined.

include/aos/common/uuid.hpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Converts UUID to string.
+ *
+ * @param uuid uuid.
+ * @param[out] dst result string.
+ * @return Error.
+ */
+inline Error UUIDToString(const Array<uint8_t>& uuid, String& dst)
```
al1img, line 68: I guess here is better to use `UUID` type.

include/aos/common/uuid.hpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Converts string to UUID.
+ *
+ * @param src input string.
+ * @param[out] dst result uuid.
+ * @return Error.
+ */
+inline Error StringToUUID(const String& src, Array<uint8_t>& dst)
```
al1img, line 83: `RetWithError`

include/aos/common/tools/error.hpp
```diff
@@ -273,6 +273,72 @@ using ErrorEnum = Error::Enum;
...
+    }
+
+    /**
+     * Constructs return value with error instance.
+     *
+     * @param value return value.
+     * @param error return error.
+     */
+    RetWithError(T&& value, const Error& error)
+        : mValue(Move(value))
```
al1img, line 295: Where is `Move` implementation?

include/aos/common/tools/error.hpp
```diff
@@ -273,6 +273,72 @@ using ErrorEnum = Error::Enum;
...
+     */
+    Error mError;
+};
+
+/**
+ * Specialization for references.
+ *
+ * @tparam T value type.
+ */
+template <typename T>
```
al1img, line 340: Why this second variant with reference is required? It is dangerous IMHO.

include/aos/common/serialnum.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include "aos/common/config.hpp"
+#include "aos/common/tools/string.hpp"
+
+namespace aos {
+namespace serialnum {
+
+/**
+ *  Serial number size(in bytes).
+ */
+constexpr auto cSerialNumLen = AOS_CONFIG_SERIAL_NUM_SIZE;
```
al1img, line 20: Don't think we need this common class for serial numbers. It looks like it will be just hex to string and string to hex converter. So, it should be part of string converters.

include/aos/common/pkcs11utils.hpp
```diff
@@ -0,0 +1,155 @@
...
+     **/
+    eP384
+};
+
+/**
+ * Crypto private key extended with object handle.
+ */
+class PrivateKey : public crypto::PrivateKey {
+public:
+    virtual ObjectHandle getHandle() = 0;
```
al1img: Should not be virtual I guess. `GetHandle` + comment.

include/aos/common/pkcs11utils.hpp
```diff
@@ -0,0 +1,155 @@
...
+ */
+class Utils {
+public:
+    /**
+     * Creates an object instance.
+     *
+     * @param session session context.
+     * @param allocator allocator for token/session objects.
+     */
+    Utils(pkcs11::SessionContext& session, Allocator& allocator)
```
al1img, line 46: Unclear why we should pass the allocator for utils. Could utils have own private allocator? I need usage example.

include/aos/common/pkcs11.hpp
```diff
@@ -0,0 +1,627 @@
...
+    Version mHardwareVersion;
+    /**
+     * Version of the slot's firmware.
+     */
+    Version mFirmwareVersion;
+
+    /**
+     * Slot flags.
+     */
+    struct Flags {
```
al1img, line 146: Why not use just enum?

include/aos/common/pkcs11.hpp
```diff
@@ -0,0 +1,627 @@
...
+        /**
+         * Flag is set if a token is present in the slot.
+         */
+        static constexpr uint32_t cTokenPresent = 0x1u;
```
al1img, line 150: Why do you need suffix `u`?

include/aos/common/pkcs11.hpp
```diff
@@ -0,0 +1,627 @@
...
+};
+
+/**
+ * The types of Cryptoki users.
+ */
+enum class UserType : uint32_t {
+    /**
+     * Security Officer.
+     */
+    eSO = 0x0u,
```
al1img, line 176: Why do you need suffix `u`?

include/aos/common/pkcs11.hpp
```diff
@@ -0,0 +1,627 @@
...
+
+/**
+ * HSM device model length.
+ */
+constexpr auto cModelLen = AOS_CONFIG_CRYPTO_PKCS11_MODEL_LEN;
+
+/**
+ * Maximum length of user PIN (password).
+ */
+static constexpr auto cPINLength = AOS_CONFIG_CRYPTO_PKCS11_PIN_LEN;
```
al1img, line 44: Why static?

include/aos/common/pkcs11utils.hpp
```diff
@@ -0,0 +1,155 @@
...
+     * @param label key label.
+     * @return RetWithError<SharedPtr<crypto::crypto::PrivateKey>>.
+     */
+    RetWithError<PrivateKey> FindPrivateKey(const Array<uint8_t>& id, const String& label)
+    {
+        (void)id;
+        (void)label;
+
+        return {PrivateKey()};
+    }
```
al1img, line 148: I guess we need something like `FindCertificateChain` as well.

include/aos/common/pkcs11utils.hpp
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright (C) 2023 Renesas Electronics Corporation.
+ * Copyright (C) 2023 EPAM Systems, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef AOS_IAM_MODULES_PKCS11_UTILS_HPP_
```
al1img, line 8: wrong name

include/aos/common/pkcs11utils.hpp
```diff
@@ -0,0 +1,155 @@
...
+#define AOS_IAM_MODULES_PKCS11_UTILS_HPP_
+
+#include "aos/common/pkcs11.hpp"
```
al1img, line 11: Could we move utils inside `pkcs11.hpp`? Otherwise, create a separate folder: `pkcs11`: and put all pkcs11-related stuff there.

include/aos/common/config.hpp
```diff
@@ -225,4 +218,81 @@
...
+ * DER certificate size.
+ */
+#ifndef AOS_CONFIG_CRYPTO_DER_CERT_SIZE
+#define AOS_CONFIG_CRYPTO_DER_CERT_SIZE 2048
+#endif
+
+/**
+ * Maximum length of PKCS11 slot description.
+ */
+#ifndef AOS_CONFIG_CRYPTO_PKCS11_SLOT_DESCRIPTION_LEN
```
al1img, line 231: All PKCS11 consts should have PKCS11 prefix not CRYPTO

include/aos/common/pkcs11/impl/pkcs11.h
```diff
@@ -0,0 +1,1526 @@
...
+   need for ABI compatibility.
+
+   CK_FALSE, CK_TRUE and NULL_PTR are removed without substitute.  Use
+   <stdbool.h>.
+
+   If CRYPTOKI_COMPAT is defined before including this header file,
+   then none of the API changes above take place, and the API is the
+   one defined by the PKCS #11 standard.  */
+
+#ifndef PKCS11_H
```
al1img, line 53: `impl` folder sounds confusing. What about `cryptoki`?

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,49 @@ class String : public Array<char> {
...
+    {
+        if (arr.MaxSize() * 2 < Size()) {
+            return Error::Enum::eNoMemory;
+        }
+
+        static const auto parseHex = [](char ch) {
+            if (ch >= '0' && ch <= '9') {
+                return static_cast<uint8_t>(ch - '0');
+            }
+            if (ch >= 'A' && ch <= 'F') {
```
al1img, line 236: Empty line between if

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,49 @@ class String : public Array<char> {
...
+                return static_cast<uint8_t>(ch - 'A' + 10);
+            }
+            return static_cast<uint8_t>(255);
```
al1img, line 239: empty line before return.

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,49 @@ class String : public Array<char> {
...
+        };
+
+        for (size_t i = 0; i <= Size(); i += 2) {
+            uint8_t low = parseHex(Get()[i]);
+            if (low > 0xFU) {
+                return AOS_ERROR_WRAP(Error::Enum::eInvalidArgument);
+            }
+            if (i + 1 > Size()) {
```
al1img, line 247: ditto

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,49 @@ class String : public Array<char> {
...
+            if (ch >= 'A' && ch <= 'F') {
+                return static_cast<uint8_t>(ch - 'A' + 10);
+            }
+            return static_cast<uint8_t>(255);
+        };
+
+        for (size_t i = 0; i <= Size(); i += 2) {
+            uint8_t low = parseHex(Get()[i]);
+            if (low > 0xFU) {
+                return AOS_ERROR_WRAP(Error::Enum::eInvalidArgument);
```
al1img, line 245: We don't wrap error in lib part: better to know function that called conversion which leads to error.

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,49 @@ class String : public Array<char> {
...
+            }
+            if (i + 1 > Size()) {
+                PushBack(low);
+                break;
+            }
+
+            uint8_t high = parseHex(Get()[i]);
+            if (high > 0xFU) {
+                return AOS_ERROR_WRAP(Error::Enum::eInvalidArgument);
```
al1img, line 254: Ditto

include/aos/common/tools/string.hpp
```diff
@@ -242,6 +285,32 @@ class String : public Array<char> {
...
+        }
+
+        Clear();
+
+        constexpr char cDigits[] = "0123456789ABCDEF";
+
+        for (const auto val : arr) {
+            const auto low = val & 0xF;
+            const auto high = (val >> 4);
+            PushBack(cDigits[low]);
```
al1img, line 307: empty line after vars definition.

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,653 @@
...
+    /**
+     * A Curve which implements NIST P-384 (FIPS 186-3, section D.2.4), also known as secp384r1
+     **/
+    eP384
+};
+
+/**
+ * Crypto private key extended with object handle.
+ */
+struct PrivateKey {
```
al1img, line 523: PrivateKey should be the class that implements crypto private key interface.

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    /**
+     * Retrieves a previously created asymmetric key pair.
+     *
+     * @param id key id.
+     * @param label key label.
+     * @return RetWithError<PrivateKey>.
+     */
+    RetWithError<PrivateKey> FindPrivateKey(const Array<uint8_t>& id, const String& label)
```
al1img, line 641: Previously it returned shared pointer to private key. What is changed?

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,653 @@
...
+ */
+class Utils {
+public:
+    /**
+     * Creates an object instance.
+     *
+     * @param session session context.
+     * @param allocator allocator for token/session objects.
+     */
+    Utils(pkcs11::SessionContext& session, Allocator& allocator)
```
al1img, line 545: Sorry, I didn't find the example of `Utils` usage. I'm trying to understand allocator usage. Let's discuss again.

include/aos/iam/modules/pkcs11.hpp
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * Module
+     */
+    bool mModulePathInURL;
+};
+
+/**
+ * A helper object for search operations containing the most valuable data for certhandler.
+ */
+struct SearchObject;
```
al1img, line 71: Looks like it is private for `PKCS11Module`. Move its definition under private section of `PKCS11Module`.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+Error PKCS11Module::Init(
+    pkcs11::LibraryContext& pkcs11, crypto::x509::ProviderItf& x509Provider, uuid::UUIDManagerItf& uuidManager)
+{
+    mPKCS11 = &pkcs11;
+    mX509Provider = &x509Provider;
+    mUUIDManager = &uuidManager;
+
+    Error err = Error::Enum::eNone;
+    Tie(cTeeClientUUIDNs, err) = mUUIDManager->StringToUUID("58AC9CA0-2086-4683-A1B8-EC4BC08E01B6");
```
al1img, line 60: Empty line after var definition.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+}
+
+Error PKCS11Module::Init(
+    pkcs11::LibraryContext& pkcs11, crypto::x509::ProviderItf& x509Provider, uuid::UUIDManagerItf& uuidManager)
+{
+    mPKCS11 = &pkcs11;
+    mX509Provider = &x509Provider;
+    mUUIDManager = &uuidManager;
+
+    Error err = Error::Enum::eNone;
```
al1img, line 59: `Error` is initialized with none: `Error err;` should be ok. But explicit initialization is ok too.

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,711 @@
...
+ */
+class LibraryContext {
+public:
+    /**
+     * Initializes PKCS11 library.
+     *
+     * @param library underlying PKCS11 C library path
+     * @return Error
+     */
+    Error InitLibrary(const String& library)
```
al1img, line 435: Take into account we may have different modules that use the same pkcs11 library. Need to discuss.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,952 @@
...
+        if (!err.IsNone()) {
+            return err;
+        }
+
+        err = CreateCurrentContext();
+    } else {
+        LOG_DBG() << "No owned token found";
+    }
+
+    return err;
```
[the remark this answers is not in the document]
al1img: I mean explicitly `return aos::ErrorEnum::eNone;`. It increases readability.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+            }
+
+            err = FS::WriteFile(mConfig.mUserPINPath, userPIN, 0600);
+            if (!err.IsNone()) {
+                return err;
+            }
+        }
+    }
+
+    LOG_DBG() << "Init token. SlotID=" << mSlotID << ", label=" << mConfig.mTokenLabel;
```
al1img, line 139: Let's have the following log format: `message: var1 = val1, var2 = var2`

`Init token: slotID = ..., label = ...`

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    UniquePtr<pkcs11::SessionContext> session;
+
+    Tie(session, err) = CreateSession(false, password);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    if (!mTeeLoginType.IsEmpty()) {
+        LOG_DBG() << "Init PIN. pin=" << mUserPIN << ", session=" << session->GetHandle();
```
al1img, line 154: ditto

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+    } else {
+        LOG_DBG() << "Init PIN. session=" << session->GetHandle();
```
al1img, line 156: ditto

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+    StaticArray<SearchObject, cCertsPerModule * 3> tokens; // certs, privKeys, pubKeys
+    SearchObject                                   filter;
+
+    err = FindObject(*session, filter, tokens);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    for (const auto& token : tokens) {
+        Error releaseErr = session->DestroyObject(token.mHandle);
```
al1img, line 192: `auto releaseErr`

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    return err;
+}
+
+RetWithError<SharedPtr<crypto::PrivateKey>> PKCS11Module::CreateKey(const String& password, KeyGenAlgorithm algorithm)
+{
+    (void)password;
+
+    PKCS11Module::PendingKey pendingKey;
+
```
al1img, line 207: remove empty line

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+RetWithError<pkcs11::SlotID> PKCS11Module::GetSlotID()
+{
+    const int paramCount
+        = static_cast<int>(mConfig.mSlotID.HasValue() + mConfig.mSlotIndex.HasValue() + !mConfig.mTokenLabel.IsEmpty());
+
+    if (paramCount > 1) {
+        LOG_ERR()
+            << "Only one parameter for slot identification should be specified (slotId or slotIndex or tokenLabel)";
+
+        return {0U, AOS_ERROR_WRAP(Error::Enum::eInvalidArgument)};
```
al1img, line 456: Why do you need `U` suffix? Do you see any warning? This question relates to all other cases as well.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+            }
+        }
+    }
+
+    if (freeSlotId.HasValue()) {
+        return {freeSlotId.GetValue(), Error::Enum::eNone};
+    }
+
+    LOG_ERR() << "No suitable slot found";
+    return {0U, Error::Enum::eNotFound};
```
al1img, line 514: Empty line before return.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+{
+    pkcs11::TokenInfo tokenInfo;
+
+    auto err = mPKCS11->GetTokenInfo(mSlotID, tokenInfo);
+    if (!err.IsNone()) {
+        return {false, err};
+    }
+
+    bool isOwned = (tokenInfo.mFlags & CKF_TOKEN_INITIALIZED) != 0;
+    return {isOwned, Error::Enum::eNone};
```
al1img, line 527: ditto or you can combine these two lines.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+    pkcs11::LibInfo   libInfo;
+    pkcs11::SlotInfo  slotInfo;
+    pkcs11::TokenInfo tokenInfo;
+
+    auto err = mPKCS11->GetLibInfo(libInfo);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    LOG_DBG() << "Library=" << mConfig.mLibrary << ", info=" << libInfo;
```
al1img, line 541: spaces befor `=`

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    err = mPKCS11->GetSlotInfo(slotId, slotInfo);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    LOG_DBG() << "SlotID=" << slotId << ", slotInfo=" << slotInfo;
```
al1img, line 548: Ditto

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+    }
+
+    return {Move(session), Error::Enum::eNone};
+}
+
+Error PKCS11Module::FindObject(pkcs11::SessionContext& session, const SearchObject& filter, Array<SearchObject>& dst)
+{
+    static constexpr auto cSearchObjAttrCount = 4U;
+
+    CK_BBOOL token = CK_TRUE;
```
al1img, line 662: `auto`?

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+}
+
+Error PKCS11Module::CreateURL(const String& label, const Array<uint8_t>& id, String& url)
+{
+    const auto addParam = [](const char* name, const char* param, bool opaque, String& paramList) {
+        if (!paramList.IsEmpty()) {
+            const char* delim = opaque ? ";" : "&";
+            paramList.Append(delim);
+        }
+        paramList.Append(name).Append("=").Append(param);
```
al1img, line 787: Empy line after `if`

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    return result;
+}
+
+Error PKCS11Module::GetX509Cert(
+    pkcs11::SessionContext& session, pkcs11::ObjectHandle object, crypto::x509::Certificate& cert)
+{
+    static constexpr auto cSingleAttribute = 1;
+
+    StaticArray<uint8_t, crypto::cDERCertSize> certBuffer;
```
al1img, line 925: Too big for stack. Allocate using allocator.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: I would keep the type explicit in case if left / right sides are of different types.

include/aos/common/tools/error.hpp
```diff
@@ -374,6 +374,79 @@ struct RetWithError<T&> {
     Error mError;
 };
 
+/**
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

include/aos/common/tools/optional.hpp
```diff
@@ -0,0 +1,135 @@
...
+ * @tparam T
+ */
+template <class T>
+class Optional {
+public:
+    /**
+     * Default constructor.
+     */
+    Optional() = default;
+    /**
```
al1img, line 25: Empty lines between functions.

include/aos/common/tools/optional.hpp
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    /**
+     * Assigns value.
+     *
+     * @param value
+     * @return void.
+     */
+    template <typename... Args>
```
al1img, line 100: Why do you need this method if it is not used?

include/aos/common/tools/optional.hpp
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Assigns value.
+     *
+     * @param value
+     * @return void.
+     */
+    void SetValue(const T& value)
+    {
+        if (HasValue()) {
+            *reinterpret_cast<T*>(mBuffer) = value;
```
al1img, line 87: Usage of `reinterpret_cast` should be minimized. Avoid it here.

include/aos/common/tools/error.hpp
```diff
@@ -279,24 +279,48 @@ struct RetWithError {
      * @param value return value.
      * @param error return error.
      */
-    RetWithError(T value, const Error& error)
+    RetWithError(const T& value, const Error& error)
```
al1img, line 282: And again: it should be two separate commits: make RetWithError to handle UniquePtr and add Tie

include/aos/common/tools/error.hpp
```diff
@@ -308,6 +332,121 @@ struct RetWithError {
...
+        return *this;
+    }
+
+    /**
+     * Assignment operator.
+     *
+     * @param src return value.
+     * @return TieWrapper<T>&
+     */
+    template <typename T1>
```
al1img, line 417: ......

tests/common/src/tools/error_test.cpp
```diff
@@ -90,3 +90,14 @@ TEST(CommonTest, ErrorMessages)
     EXPECT_EQ(Error(ErrorEnum::eAlreadyExist).Message(), "already exist");
     EXPECT_EQ(Error(ErrorEnum::eWrongState).Message(), "wrong state");
 }
+
+TEST(CommonTest, ErrorTie)
+{
+    Error err = Error::Enum::eNone;
```
al1img, line 96: Use ErrorEnum everywhere.

include/aos/iam/certhandler.hpp
```diff
@@ -9,6 +9,8 @@
 #define AOS_CERTHANDLER_HPP_
 
 #include "aos/common/tools/thread.hpp"
+
+#include "aos/iam/config.hpp"
```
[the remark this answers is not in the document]
al1img: I'm asking about HSMItf and CertInfo?

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    /**
+     * Converts byte array to string.
+     *
+     * @param src source byte array.
+     * @param dst destination string.
+     * @return Error.
+     */
+    static Error Convert(const Array<uint8_t>& src, const String& dst)
```
al1img, line 672: Don't understand why do you need this method.

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    /**
+     * Converts string to byte array.
+     *
+     * @param src source string.
+     * @param dst destination byte array.
+     * @return Error.
+     */
+    static Error Convert(const String& src, const Array<uint8_t>& dst)
```
al1img, line 687: ditto.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,995 @@
...
+    if (isOwned) {
+        err = PrintInfo(mSlotID);
+        if (!err.IsNone()) {
+            return err;
+        }
+    } else {
+        LOG_DBG() << "No owned token found";
+    }
+
+    return Error::Enum::eNone;
```
al1img, line 102: Use `ErrorEnum`

include/aos/iam/modules/pkcs11.hpp
```diff
@@ -0,0 +1,237 @@
...
+    Error GeneratePIN(const String& loginType, const uuid::UUID& space, uint32_t data, String& userPIN);
+    Error GetUserPin(String& pin) const;
+
+    RetWithError<UniquePtr<pkcs11::SessionContext>> CreateSession(bool userLogin, const String& pin);
+
+    Error FindObject(pkcs11::SessionContext& session, const SearchObject& filter, Array<SearchObject>& dst);
+
+    Error TokenMemInfo() const;
+
+    static bool CheckCertificate(const crypto::x509::Certificate& cert, const crypto::PrivateKey& key);
```
al1img, line 200: Why do you need it static?

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,995 @@
...
+    err = FindObject(*session, filter, tokens);
+    if (!err.IsNone()) {
+        return err;
+    }
+
+    for (const auto& token : tokens) {
+        auto releaseErr = session->DestroyObject(token.mHandle);
+        if (!releaseErr.IsNone()) {
+            err = releaseErr;
+            LOG_ERR() << "Can't delete object: Handle = " << token.mHandle;
```
al1img, line 196: `handle = `

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,995 @@
...
+        return {nullptr, AOS_ERROR_WRAP(err)};
+    }
+
+    switch (algorithm.GetValue()) {
+    case KeyGenAlgorithmEnum::eRSA:
+        Tie(pendingKey.mKey, err) = pkcs11::Utils(*session, mLocalCacheAllocator)
+                                        .GenerateRSAKeyPairWithLabel(pendingKey.mUUID, mCertType, cRSAKeyLength);
+        if (!err.IsNone()) {
+            return {nullptr, AOS_ERROR_WRAP(err)};
+        }
```
al1img, line 228: empty line after if.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,995 @@
...
+        break;
+
+    case KeyGenAlgorithmEnum::eECC:
+        Tie(pendingKey.mKey, err) = pkcs11::Utils(*session, mLocalCacheAllocator)
+                                        .GenerateECDSAKeyPairWithLabel(pendingKey.mUUID, mCertType, cECSDACurveID);
+        if (!err.IsNone()) {
+            return {nullptr, AOS_ERROR_WRAP(err)};
+        }
```
al1img, line 236: ditto

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    /**
+     * Retrieves a previously created asymmetric key pair.
+     *
+     * @param id key id.
+     * @param label key label.
+     * @return RetWithError<SharedPtr<PrivateKey>>.
+     */
+    RetWithError<SharedPtr<PrivateKey>> FindPrivateKey(const Array<uint8_t>& id, const String& label)
```
al1img, line 657: What would happen if utils and session will be destroyed? Will this key works as expected?

include/aos/common/tools/string.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include <ctype.h>
 #include <inttypes.h>
+#include <regex.h>
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

include/aos/common/tools/string.hpp
```diff
@@ -331,6 +423,46 @@ class String : public Array<char> {
...
+
+    /**
+     * Executes search by specififed regex string and returns match by its number.
+     *
+     * @tparam MatchInd specifies index of the match to be returned.
+     * @param regex POSIX extended regular expression.
+     * @param[out] match result match.
+     * @return Error.
+     */
+    template <int MatchInd>
```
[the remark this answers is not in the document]
mykola-kobets-epam: fixed

include/aos/common/pkcs11/pkcs11.hpp
mykola-kobets-epam: Yes, it is.  SharedPtr is allocated with external allocator. Produced handles are not bound to one session.

include/aos/common/pkcs11/pkcs11.hpp
mykola-kobets-epam: In order to convert Label to and from pkcs11 UTF8 string(which is a byter array). It is needed because certmodule pkcs11 adds custom SearchObject with label & uses low level FindObject/GetObjAttrib with it.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: Still not clear why ErrorEnum is needed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed

include/aos/iam/modules/pkcs11.hpp
mykola-kobets-epam: Basically that's only to say that method doesnt require context of an object.
But I will remove static to make prototype uniform.

include/aos/iam/modules/config.hpp
```diff
@@ -0,0 +1,53 @@
+/*
```
mykola-kobets-epam, line 1: Missing discussion:

No need to move cert modules options into separate file. Except ones, directly related to pkcs11 module. Also, we should have possibility to enable/disable this module compile time. So, I think all pkcs11 module files should be move into separate pkcs11 module folder.
Author
@[mykola-kobets-epam](https://github.com/mykola-kobets-epam) mykola-kobets-epam [Dec 8, 2023](https://github.com/aoscloud/aos_core_lib_cpp/pull/47#discussion_r1420296255)

    No need to move cert module options into separate file
    In this case it's not clear where all of this options should be located in. Also I want to minimize dependencies from the upper level folders within a module (as its hard to track interrelationships within a module when you have header files included from upper and lower dir level). Currently I'm trying to add all necessary data closer to the context of use.

    Also, we should have possibility to enable/disable this module compile time.
    I think it would better to implement this option when all of the related parts are implemented. Probably I'll add it in cryproki wrapper PR.

    pkcs11 module files should be move into separate pkcs11 module folder
    Added pkcs11 folder, moved some config options there.

tests/common/src/tools/string_test.cpp
```diff
@@ -157,3 +157,46 @@ TEST(CommonTest, SplitString)
...
+    StaticArray<uint8_t, 8> result;
+    uint8_t                 expected[] = {0xAB, 0xCD, 0xEF, 0x01, 0x23, 0x45, 0x67, 0x89};
+
+    ASSERT_TRUE(hex.ToByteArr(result).IsNone());
+    EXPECT_EQ(result, Array<uint8_t>(expected, sizeof(expected)));
+}
+
+TEST(CommonTest, StringToByteArrayOddSize)
+{
+    const String            hex = "01234";
```
al1img, line 174: Empty line after const?

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,56 @@ class String : public Array<char> {
...
+                return static_cast<uint8_t>(ch - 'A' + 10);
+            }
+
+            return static_cast<uint8_t>(255);
+        };
+
+        arr.Clear();
+
+        for (size_t i = 0; i < Size(); i += 2) {
+            uint8_t high = parseHex(Get()[i]);
```
al1img, line 252: You can use `(*this)[i]` instead of `Get` as it checks the range.

include/aos/common/tools/string.hpp
```diff
@@ -218,6 +218,56 @@ class String : public Array<char> {
...
+            if (high > 0xFU) {
+                return Error::Enum::eInvalidArgument;
+            }
+
+            if (i + 1 >= Size()) {
+                arr.PushBack(high << 4);
+                break;
+            }
+
+            uint8_t low = parseHex(Get()[i + 1]);
```
al1img, line 262: ditto

include/aos/common/config.hpp
```diff
@@ -239,4 +239,60 @@
...
+
+/**
+ * Maximum length of PKCS11 manufacture ID.
+ */
+#ifndef AOS_CONFIG_PKCS11_MANUFACTURE_ID_LEN
+#define AOS_CONFIG_PKCS11_MANUFACTURE_ID_LEN 32
+#endif
+
+/**
+ * Maximum length of PKCS11 lable.
```
al1img, line 257: label

include/aos/common/pkcs11/pkcs11.hpp
al1img: Is it just cast aks uint8_t[] to char[]? Or uint8_t[] to string and back?

If it is first case then it should be handled in String and Array classes. We should have straight simple conversion between Array<uint8_t> and String. It would be useful in other cases as well.

For the second case it is already done on String class.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
+/*
```
al1img, line 1: We should move cpp file to separate `pkcs11` folder as well. Similar to `pkcs11.hpp`.

src/iam/certhandler/modules/pkcs11.cpp
al1img: Probably yes for Error::Enum to avoid ambiguity. But it is done in this way for all Aos enums:
```c++
using ExtendedKeyUsageEnum = ExtendedKeyUsageType::Enum;
using ExtendedKeyUsage = EnumStringer<ExtendedKeyUsageType>;
```
I don't want to make the exception for Error.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+        AOS_ERROR_WRAP(err);
+    }
+
+    Tie(sha1Str, err) = mUUIDManager->UUIDToString(sha1);
+    if (!err.IsNone()) {
+        AOS_ERROR_WRAP(err);
+    }
+
+    return userPIN.Format("%s:%s", loginType.CStr(), sha1Str.CStr());
+    ;
```
al1img, line 603: remove

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+        pin.Clear();
+        return Error::Enum::eNone;
+    }
+
+    return FS::ReadFile(mConfig.mUserPINPath, pin);
+}
+
+RetWithError<UniquePtr<pkcs11::SessionContext>> PKCS11Module::CreateSession(bool userLogin, const String& pin)
+{
+    using pkcs11::SessionInfo;
```
al1img, line 618: Why do you need it? SessionInfo used only once in this function.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+    static constexpr auto cSearchObjAttrCount = 4U;
+
+    // create search template
+    CK_BBOOL token = CK_TRUE;
+
+    StaticArray<pkcs11::ObjectAttribute, cSearchObjAttrCount> templ;
+    templ.EmplaceBack(CKA_TOKEN, Array<uint8_t>(&token, sizeof(token)));
+
+    if (!filter.mID.IsEmpty()) {
+        templ.EmplaceBack(CKA_ID, Array<uint8_t>(filter.mID.Get(), filter.mID.Size()));
```
al1img, line 674: mID is array itself

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+    }
+
+    if (!filter.mLabel.IsEmpty()) {
+        const auto labelPtr = reinterpret_cast<const uint8_t*>(filter.mLabel.Get());
```
al1img, line 678: Why intermediate var is needed? And if we implement implicit Array<uint8_t> <-> String conversion this cast will not be needed.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+        templ.EmplaceBack(CKA_LABEL, Array<uint8_t>(labelPtr, filter.mLabel.Size()));
+    }
+
+    if (!filter.mType.HasValue()) {
+        const auto classPtr = reinterpret_cast<const uint8_t*>(&filter.mType.GetValue());
```
al1img, line 683: ditto

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+
+Error PKCS11Module::FindObject(pkcs11::SessionContext& session, const SearchObject& filter, Array<SearchObject>& dst)
+{
+    static constexpr auto cSearchObjAttrCount = 4U;
+
+    // create search template
+    CK_BBOOL token = CK_TRUE;
+
+    StaticArray<pkcs11::ObjectAttribute, cSearchObjAttrCount> templ;
+    templ.EmplaceBack(CKA_TOKEN, Array<uint8_t>(&token, sizeof(token)));
```
al1img, line 671: Empty line after var definition.

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+    // search object handles
+    StaticArray<pkcs11::ObjectHandle, cCertsPerModule * 3> objects; // certs, privKeys, pubKeys
+
+    auto err = session.FindObjects(templ, objects);
+    if (!err.IsNone()) {
+        return AOS_ERROR_WRAP(err);
+    }
+
+    // retrieve attributes(id & label) and add search objects
+    StaticArray<pkcs11::AttributeType, cSearchObjAttrCount> searchAttrTypes;
```
al1img, line 696: empty line after var definition

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+    }
+
+    return Error::Enum::eNone;
+}
+
+Error PKCS11Module::TokenMemInfo() const
+{
+    pkcs11::TokenInfo info;
+
+    Error err = mPKCS11->GetTokenInfo(mSlotID, info);
```
al1img, line 735: auto err

src/iam/certhandler/modules/pkcs11.cpp
```diff
@@ -0,0 +1,997 @@
...
+}
+
+Error PKCS11Module::GetX509Cert(
+    pkcs11::SessionContext& session, pkcs11::ObjectHandle object, crypto::x509::Certificate& cert)
+{
+    static constexpr auto cSingleAttribute = 1;
+
+    auto certBuffer = MakeUnique<DERCert>(&mTmpObjAllocator);
+
+    StaticArray<pkcs11::AttributeType, cSingleAttribute> types;
```
al1img, line 938: Empty line after var definition and combine var block:
```c++
    StaticArray<pkcs11::AttributeType, cSingleAttribute> types;
    StaticArray<Array<uint8_t>, cSingleAttribute> values;

    types.PushBack(CKA_VALUE);
    values.PushBack(*certBuffer);
```

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed.

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: done in a separate commit

src/iam/certhandler/modules/pkcs11.cpp
mykola-kobets-epam: fixed

include/aos/common/tools/fs.hpp
```diff
@@ -348,6 +348,38 @@ class FS {
...
     }
+
+    /**
+     * Reads content of the file named by fileName into the given string.
+     *
+     * @param fileName file name.
+     * @param[out] text result string.
+     * @return Error.
+     */
+    static Error ReadFile(const String& fileName, String& text)
```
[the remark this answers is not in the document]
al1img: Your C++ example more related how stream handle lines in text file rather than deal with string. How do you read whole file into std::string with one method in C++?

Also here is golang API:
```go
func ReadFile(name string) []byte
``` 

My consern here, you can't use this API for reading binary files in case we need it in the future. And we will need to implement separate methods for such case. Having byte array covers both cases and we may use one implementation for both.

But if you really like to read direct to string, then these methods should be renamed to: `ReadFileToString` and `WriteStringToFile` accordingly.

include/aos/common/crypto.hpp
```diff
@@ -83,17 +79,184 @@ constexpr auto cSerialNumSize = AOS_CONFIG_CRYPTO_SERIAL_NUM_SIZE;
...
+
+/**
+ * Supported key types.
+ */
+enum class KeyType { eRSA, eECDSA };
+
+/**
+ * Public key interface.
+ */
+class PublicKey {
```
al1img, line 109: Add Itf suffix for all interfaces. I missed it in the previous PR, so please add it for PrivateKey interface as well.

include/aos/common/crypto.hpp
```diff
@@ -83,17 +79,184 @@ constexpr auto cSerialNumSize = AOS_CONFIG_CRYPTO_SERIAL_NUM_SIZE;
...
+     * @param e public exponent.
+     */
+    RSAPublicKey(const Array<uint8_t>& n, const Array<uint8_t>& e);
+
+    /**
+     * Returns type of a public key.
+     *
+     * @return KeyType,
+     */
+    virtual KeyType GetKeyType() const;
```
al1img, line 168: add override keyword

include/aos/common/crypto.hpp
```diff
@@ -83,17 +79,184 @@ constexpr auto cSerialNumSize = AOS_CONFIG_CRYPTO_SERIAL_NUM_SIZE;
...
+
+    /**
+     * Tests whether current key is equal to the provided one.
+     *
+     * @param pubKey public key.
+     * @return bool.
+     */
+    virtual bool IsEqual(const PublicKey& pubKey) const;
```
al1img, line 176: ditto

include/aos/common/crypto.hpp
al1img: add implementation here. In this case const getters usually automatically inlined AFAIK.

include/aos/common/crypto.hpp
```diff
@@ -83,17 +79,184 @@ constexpr auto cSerialNumSize = AOS_CONFIG_CRYPTO_SERIAL_NUM_SIZE;
...
+ * RSA private key.
+ */
+class RSAPrivateKey : public PrivateKey {
+public:
+    /**
+     * Constructs object instance.
+     *
+     * @param pubKey public key component of a private key.
+     */
+    RSAPrivateKey(const RSAPublicKey& pubKey);
```
al1img, line 193: Ditto

src/common/crypto.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+/***********************************************************************************************************************
+ * Public
+ **********************************************************************************************************************/
+
+//
+// RSAPublicKey implementation
+//
+
+RSAPublicKey::RSAPublicKey(const Array<uint8_t>& n, const Array<uint8_t>& e)
```
al1img, line 21: Remove cpp file and put implementation directly into header file. These are mostly const methods. Also we initilly designed this lib to be header only. It is not always possible. But we try to keep it in this format.

include/aos/common/crypto.hpp
mykola-kobets-epam: not sure virtual functions are inlined well. but anyway I moved implementation into the header.

src/common/crypto.cpp
mykola-kobets-epam: done

include/aos/common/tools/fs.hpp
```diff
@@ -348,6 +348,38 @@ class FS {
...
+
+    /**
+     * Reads content of the file named by fileName into the given string.
+     *
+     * @param fileName file name.
+     * @param[out] text result string.
+     * @return Error.
+     */
+    static Error ReadFileToString(const String& fileName, String& text)
+    {
```
mykola-kobets-epam, line 360: restore the discussion:

@[al1img](https://github.com/al1img) al1img [Dec 15, 2023](https://github.com/aoscloud/aos_core_lib_cpp/pull/47#discussion_r1428076393)

Your C++ example more related how stream handle lines in text file rather than deal with string. How do you read whole file into std::string with one method in C++?

Also here is golang API:

func ReadFile(name string) []byte

My consern here, you can't use this API for reading binary files in case we need it in the future. And we will need to implement separate methods for such case. Having byte array covers both cases and we may use one implementation for both.

But if you really like to read direct to string, then these methods should be renamed to: ReadFileToString and WriteStringToFile accordingly.
Author
@[mykola-kobets-epam](https://github.com/mykola-kobets-epam) mykola-kobets-epam [Dec 15, 2023](https://github.com/aoscloud/aos_core_lib_cpp/pull/47#discussion_r1428160966)

    How do you read whole file into std::string with one method in C++?

One example I have found:
string get_file_string(){
std::ifstream ifs("path_to_file");
return string((std::istreambuf_iterator(ifs)), (std::istreambuf_iterator()));
}

    Having byte array covers both cases and we may use one implementation for both.

As I see golang has single readFile method to read content into a byte array. And conversion constructor to convert byte array into a string. So basically we need two methods as well.

There are similary apis in other languages. example for java:
Java 7
String content = new String(Files.readAllBytes(Paths.get("readMe.txt")), StandardCharsets.UTF_8);
Java 11
String content = Files.readString(Paths.get("readMe.txt"));

I chose the latest approach as it is more convenient,
otherwise it would be needed to allocate additional byte array, read content there & make conversion to string.

    these methods should be renamed to: ReadFileToString and WriteStringToFile accordingly.

ok, I will rename them.

al1img: In all provided by you examples, byte array converted to string.

include/aos/common/pkcs11/pkcs11.hpp
```diff
@@ -0,0 +1,743 @@
...
+    }
+
+    /**
+     * Converts PKCS11 byte array to string.
+     *
+     * @param src source byte array.
+     * @param dst destination string.
+     * @return Error.
+     */
+    static Error Convert(const Array<uint8_t>& src, String& dst)
```
al1img, line 713: I've asked before but didn't find the answer. What function suppose to do? Just copy byte array to string?
If so, it is not task for pkcs11.  It should be done in common.

mykola-kobets-epam: here is an example from documentation
slotDescription       character-string description of the slot.  MUST be padded with the blank character (‘ ‘).  MUST NOT be null-terminated.

al1img (approved):

al1img (approved):

PR #48, opened 2023-11-29
Add assigning derived class smart pointers

al1img:


include/aos/common/tools/memory.hpp
```diff
@@ -44,25 +47,40 @@ class SmartPtr {
 
         assert(!object || allocator);
```
mykola-kobets-epam, line 49: dont understand the need for this assert

include/aos/common/tools/memory.hpp
```diff
@@ -190,7 +224,45 @@ class SharedPtr : public SmartPtr<T> {
      */
     SharedPtr& operator=(const SharedPtr& ptr)
     {
-        Reset(ptr.mAllocator, ptr.mObject);
+        SmartPtr<T>::Release(ptr.GetAllocator(), ptr.Get());
```
mykola-kobets-epam, line 228: probably need to use GiveAllocation after release, right?

mykola-kobets-epam (approved):

include/aos/common/tools/memory.hpp
al1img: Valid cases:

object && allocator
!object && allocator
!object && !allocator

This is short condition to cover them all.

mykola-kobets-epam (approved):

PR #53, opened 2023-12-29
Add communication

al1img:


MykolaSolyanko (approved):

include/aos/common/tools/array.hpp
```diff
@@ -642,6 +642,19 @@ class StaticArray : public Array<T> {
...
     }
 
+    /**
+     * Assigns static array from another  array.
+     *
+     * @param array array to assign from.
+     */
+    StaticArray& operator=(const Array<T>& array)
+    {
+        Array<T>::SetBuffer(mBuffer);
```
mykola-kobets-epam, line 652: It seems like buffer is not supposed to be changed since creation. So setBuffer should be called in constructors only.

include/aos/common/tools/string.hpp
```diff
@@ -187,7 +180,7 @@ class String : public Array<char> {
      * @param srt string to compare with.
      * @return bool.
      */
-    friend bool operator==(const char* cStr, const String& str) { return str.operator==(cStr); };
+    friend bool operator==(const char* cStr, const String& str) { return str == String(cStr); };
```
mykola-kobets-epam, line 183: It looks like we only need operators ==/!= (const String&, const String&) and const char* would be implicitly converted to to String to match the operand type.

mykola-kobets-epam (approved):

PR #54, opened 2024-01-02
Miscellaneous changes in common & iam components

mykola-kobets-epam:


include/aos/common/crypto.hpp
```diff
@@ -358,7 +358,7 @@ struct Certificate {
      */
     StaticArray<uint8_t, cCertIssuerSize> mIssuer;
     /**
-     * DER encoded certificate serial number.
+     * Certificate serial number.
```
al1img, line 361: Actually you should put message body as commit header. It is self-explanatory,

mykola-kobets-epam: Added additional comment concerning Certificate::mSerial type.

MykolaSolyanko (approved):

al1img (approved):

PR #55, opened 2024-01-08
Clock sync updates

al1img:


MykolaSolyanko (approved):

tests/common/src/tools/thread_test.cpp
```diff
@@ -140,6 +140,80 @@ TEST(CommonTest, CondVar)
     EXPECT_TRUE(worker.Join().IsNone());
 }
 
+TEST(CommonTest, CondVarTimeout)
```
mykola-kobets-epam, line 143: Probably it would better separate them on different test cases.
And I'm not sure we should put all the tests in CommonTest, I would make a separate ThreadTest

mykola-kobets-epam (approved):

tests/common/src/tools/thread_test.cpp
al1img: Put each common class in its own test.

mykola-kobets-epam (approved):

PR #56, opened 2024-01-09
[iam] Replace subject with commmon name in CertHandler::CreateKey

mykola-kobets-epam:


include/aos/iam/certhandler.hpp
```diff
@@ -68,13 +68,13 @@ class CertHandler {
      * Creates key pair.
      *
      * @param certType certificate type.
-     * @param subject subject of CSR.
+     * @param subject common name of the subject.
```
al1img, line 71: Add proper param name: `subjectCommonName`

include/aos/iam/modules/certmodule.hpp
```diff
@@ -191,12 +191,12 @@ class CertModule {
     /**
      * Creates certificate request.
      *
-     * @param subject certificate subject.
+     * @param subjectCommonName common name of the subject.
```
al1img, line 194: ditto

include/aos/iam/certhandler.hpp
mykola-kobets-epam: fixed

include/aos/iam/modules/certmodule.hpp
mykola-kobets-epam: param name is already correct

al1img (approved):

PR #66, opened 2024-01-29
Bump to mbedtls 3.5.0

MykolaSolyanko:


src/common/crypto/mbedtls/psa_crypto_driver_wrappers_no_static.c
```diff
@@ -0,0 +1,282 @@
+/*
```
mykola-kobets-epam, line 1: It looks like old *wrappers.c is not needed, right? If so, please remove it.

mykola-kobets-epam (approved):

PR #72, opened 2024-02-01
[crypto driver] Move driver wrapper to include

MykolaSolyanko:


al1img (approved):

include/aos/common/crypto/mbedtls/driverwrapper.hpp
```diff
@@ -0,0 +1,31 @@
+/*
```
al1img, line 1: tag in commit message: `[common,crypto]`

include/aos/common/crypto/mbedtls/driverwrapper.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * @return Error code
+ */
+aos::RetWithError<psa_key_id_t> AosPsaAddKey(const aos::crypto::PrivateKeyItf& mPrivKey);
+
+/**
+ * @brief Remove key from the list of builtin keys
+ *
+ * @param keyId Key ID
+ */
+void AosPsaRemoveKey(psa_key_id_t keyId);
```
mykola-kobets-epam, line 29: probably need to return Error for NotFound case

MykolaSolyanko: I'm not sure this is necessary, and i think when we delete and the key is not present it's not an error

mykola-kobets-epam (approved):

al1img (approved):

PR #94, opened 2024-02-17
[iam,certmodules,pkcs11] Do not return error if object not found

al1img:
On clear cert module or validate certs we return error if pkcs11 objects not found. It shouldn't be treated as errors.

### DIFF
--- a/src/iam/certmodules/pkcs11/pkcs11.cpp
+++ b/src/iam/certmodules/pkcs11/pkcs11.cpp
@@ -157,8 +157,8 @@ Error PKCS11Module::Clear()
     }
 
     // certs, privKeys, pubKeys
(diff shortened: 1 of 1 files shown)


MykolaSolyanko (approved):

src/iam/certmodules/pkcs11/pkcs11.cpp
```diff
@@ -373,21 +377,21 @@ Error PKCS11Module::ValidateCertificates(
     filter.mType  = CKO_CERTIFICATE;
 
     err = FindObject(*session, filter, certificates);
-    if (!err.IsNone()) {
+    if (!err.IsNone() && !err.Is(aos::ErrorEnum::eNotFound)) {
```
mykola-kobets-epam, line 380: aos is not required

mykola-kobets-epam (approved):

PR #106, opened 2024-03-29
[common,config] Add max number of nodes configuration

mlohvynenko:


include/aos/common/config.hpp
```diff
@@ -155,6 +155,13 @@
 #define AOS_CONFIG_TYPES_PARTITION_TYPES_LEN 32
 #endif
 
+/**
+ * Max number of nodes.
+ */
+#ifndef AOS_CONFIG_TYPES_MAX_NUM_NODES
```
al1img, line 161: Add appropriate constexpr in common/types.hpp

mlohvynenko: Done

al1img (approved):

PR #107, opened 2024-04-01
Use RFC7512 for pkcs11 ID string representation

al1img:


MykolaSolyanko (approved):

mlohvynenko (approved):

include/aos/common/tools/string.hpp
```diff
@@ -242,62 +313,33 @@ class String : public Array<char> {
...
         for (size_t i = 0; i < Size(); i += 2) {
             Error   err = ErrorEnum::eNone;
-            uint8_t high;
+            uint8_t byte;
 
-            Tie(high, err) = HexToByte((*this)[i]);
-            if (!err.IsNone()) {
-                return err;
-            }
+            char hex[2] = {(*this)[i], '0'};
```
mykola-kobets-epam, line 318: it would better add  \0
To have a correct String representation.

include/aos/common/tools/string.hpp
```diff
@@ -306,10 +348,17 @@ class String : public Array<char> {
         Clear();
 
         for (const auto val : src) {
-            const auto digits = ByteToHex(val);
+            char digits[2];
```
mykola-kobets-epam, line 351: string.ByteToHex uses resize that has *end() = 0 that results in stack corruption.
It's better use StaticString for the particular case: no need in convertion to use ByteToHex & Instead can be replaced with append.

include/aos/common/tools/string.hpp
```diff
@@ -306,10 +348,17 @@ class String : public Array<char> {
...
 
-            PushBack(digits.mFirst);
-            PushBack(digits.mSecond);
+            auto err = String(digits, 2).ByteToHex(val, upperCase);
+            if (!err.IsNone()) {
+                return err;
+            }
+
+            err = Insert(end(), digits, static_cast<char*>(digits) + 2);
```
mykola-kobets-epam, line 358: Size is tested at the start of the method. thereis no need in additional check.

src/common/cryptoutils.cpp
```diff
@@ -19,6 +19,59 @@ constexpr auto cSchemeFile      = "file";
...
+ **********************************************************************************************************************/
+
+static Error StrToPKCS11ID(const String& idStr, Array<uint8_t>& id)
+{
+    id.Clear();
+
+    auto                 percentDetected = false;
+    aos::StaticString<2> hexByte;
+
+    for (const auto& ch : idStr) {
```
mykola-kobets-epam, line 33: it's better to use for(int i=0; i<Size(); i +=3) to simplify the logic

src/common/cryptoutils.cpp
```diff
@@ -19,6 +19,59 @@ constexpr auto cSchemeFile      = "file";
 constexpr auto cSchemePKCS11    = "pkcs11";
 constexpr auto cSchemeMaxLength = Max(sizeof(cSchemeFile), sizeof(cSchemePKCS11));
 
+/***********************************************************************************************************************
+ * Static
+ **********************************************************************************************************************/
+
+static Error StrToPKCS11ID(const String& idStr, Array<uint8_t>& id)
```
mykola-kobets-epam, line 26: Not a good name: I would suggest DecodePKCS11ID

src/iam/certmodules/pkcs11/pkcs11.cpp
```diff
@@ -18,6 +18,31 @@ namespace aos {
 namespace iam {
 namespace certhandler {
 
+/***********************************************************************************************************************
+ * Static
+ **********************************************************************************************************************/
+
+static Error IDToStr(const Array<uint8_t>& id, String& idStr)
```
mykola-kobets-epam, line 25: I would suggest renaming to EncodePKCS11ID

src/iam/certmodules/pkcs11/pkcs11.cpp
```diff
@@ -18,6 +18,31 @@ namespace aos {
...
+{
+    for (auto byte : id) {
+        auto err = idStr.PushBack('%');
+        if (!err.IsNone()) {
```
mykola-kobets-epam, line 29: as we allready know the required size of destination array I would suggest to have a single check at the beginning of the method

include/aos/common/tools/string.hpp
al1img: StaticString is not available here.

include/aos/common/tools/string.hpp
al1img: I removed checking size at the beginning then. Insert may check other errors in general case not only size.

src/common/cryptoutils.cpp
al1img: done

src/common/cryptoutils.cpp
al1img: No. Current implementation is more generic and supports the following cases for example:
`id%01%12` that is not recommended but not forbidden as well.

src/iam/certmodules/pkcs11/pkcs11.cpp
al1img: You should not rely on implementation details. `PushBack` may return other than size overflow errors as well. Normally, if method returns error, it should be handled. And because `PushBack` also checks the size, no need for checking the size at the beginning.

src/iam/certmodules/pkcs11/pkcs11.cpp
al1img: done

include/aos/common/tools/string.hpp
mykola-kobets-epam: ah, right

include/aos/common/tools/string.hpp
mykola-kobets-epam: ok

mykola-kobets-epam (approved):

PR #111, opened 2024-04-11
[EPMPAOS-4466] Use unique_ptr to deallocate C style resources 

mykola-kobets-epam:


MykolaSolyanko (approved):

include/aos/common/tools/memory.hpp
```diff
@@ -441,6 +441,21 @@ inline UniquePtr<T> MakeUnique(Allocator* allocator, Args&&... args)
...
+/**
+ * Defers object destruction till the end of the current scope.
+ *
+ * @tparam T type of the object to be destroyed.
+ * @tparam Deleter type of the deleter.
+ * @param ptr pointer to the object to be destroyed.
+ * @param deleter functor object to be deferred.
+ * @return UniquePtr<T, Deleter>.
+ */
+template <typename T, typename Deleter>
```
al1img, line 453: No need for this wrapper. We should use the deleted as follows:
```c++
auto drbg = aos::UniquePtr<mbedtls_ctr_drbg_context>(&ctrDrbg, mbedtls_ctr_drbg_free);
```

tests/common/src/tools/memory_test.cpp
```diff
@@ -143,3 +143,25 @@ TEST(MemoryTest, SharedPtrDerivedClass)
 
     EXPECT_EQ(allocator.FreeSize(), allocator.MaxSize());
 }
+
+TEST(MemoryTest, DeferRelease)
+{
+    int dummy = 0x42;
```
al1img, line 149: combine vars in one block:
```c++
  int dummy = 0x42;
  testing::MockFunction<void(int*)> deleter;

...
```

tests/common/src/tools/memory_test.cpp
```diff
@@ -143,3 +143,25 @@ TEST(MemoryTest, SharedPtrDerivedClass)
...
+    testing::MockFunction<void(int*)> deleter;
+    EXPECT_CALL(deleter, Call(&dummy)).Times(1);
+    {
+        auto defer = DeferRelease(&dummy, deleter.AsStdFunction());
+    }
+}
+
+TEST(MemoryTest, DeferReleaseNotCalledForNull)
+{
+    using namespace testing;
```
al1img, line 160: Don't use using for one call only.

tests/common/src/tools/memory_test.cpp
```diff
@@ -143,3 +143,25 @@ TEST(MemoryTest, SharedPtrDerivedClass)
...
+
+    MockFunction<void(int*)> deleter;
```
al1img, line 162: Empty line after var definition

src/common/crypto/mbedtls/cryptoprovider.cpp
```diff
@@ -137,29 +137,25 @@ static Error ParsePrivateKey(const String& pemCAKey, mbedtls_pk_context& privKey
     mbedtls_entropy_context  entropy;
 
     mbedtls_ctr_drbg_init(&ctrDrbg);
+    auto freeDRBG = DeferRelease(&ctrDrbg, mbedtls_ctr_drbg_free);
```
al1img, line 140: rename `crtDrbg` to `crtDrbgCtx` for example and then use: `auto crtDrbg = `

include/aos/common/tools/memory.hpp
mykola-kobets-epam: I see gcc has some limitations on CTAD, particularary for std::unique_ptr. So, both parameters should be provided in the example above like:
auto drbg = aos::UniquePtr<mbedtls_ctr_drbg_context, void (*)(mbedtls_ctr_drbg_context*)>(&ctrDrbg, mbedtls_ctr_drbg_free); 
and it doesn't look very good.

tests/common/src/tools/memory_test.cpp
mykola-kobets-epam: done

tests/common/src/tools/memory_test.cpp
mykola-kobets-epam: I used it for two cases, but it doesn't matter much

tests/common/src/tools/memory_test.cpp
mykola-kobets-epam: fixed

include/aos/common/tools/memory.hpp
```diff
@@ -12,6 +12,59 @@
 
 namespace aos {
 
+/**
+ * Default deleter invokes delete operator for the given pointer.
+ */
+template <typename T>
+class DefaultDeleter {
```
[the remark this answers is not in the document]
mykola-kobets-epam: In general, Deleter doesn't have to be strictly movable class. I just provided a minimal number of constructors required for our purposes.  But it doesn't matter much.
Made deleter noncopiable too.

src/common/crypto/mbedtls/cryptoprovider.cpp
mykola-kobets-epam: I didn't understand the comment.

include/aos/common/tools/memory.hpp
```diff
@@ -12,6 +12,59 @@
...
+     *
+     * @param allocator input allocator.
+     */
+    DefaultDeleter(Allocator* allocator = nullptr)
+        : mAllocator(allocator)
+    {
+    }
+
+    /**
+     * Move constructor for a DefaultDeleter of a derived class.
```
al1img, line 32: Moves

include/aos/common/tools/memory.hpp
```diff
@@ -12,6 +12,59 @@
...
+     */
+    template <typename P>
+    DefaultDeleter(DefaultDeleter<P>&& other)
+    {
+        *this = Move(other);
+    }
+
+    /**
+     * Move assignment operator for a DefaultDeleter of a derived class.
```
al1img, line 41: Ditto

include/aos/common/tools/memory.hpp
```diff
@@ -122,66 +175,136 @@ class SmartPtr {
  *
  * @tparam T holding object type.
  */
-template <typename T>
-class UniquePtr : public SmartPtr<T>, private NonCopyable {
+template <typename T, typename Deleter = DefaultDeleter<T>>
```
al1img, line 178: shared_ptr also may have custom deleter, so both smart pointers should have similar functionality

include/aos/common/tools/memory.hpp
mykola-kobets-epam: Basically move is not a verb  but noun, like "copy constructor"

include/aos/common/tools/memory.hpp
mykola-kobets-epam: There is a subtle difference between deleter in  unique & shared pointers, that makes it difficult to have a common implementation. Deleter for unique_ptr is stored on the stack, while shared_ptr's deleter stored in the heap(allocator's space in our case). SmartPtr implements a shared_ptr's approach, so I made a separate impl for unique_ptr.
 
There is also a difference in behaviour for the following example:
struct A{}; struct B: A {}
unique_ptr<A> b = make_unique<B>();
shared_ptr<A> b = make_shared<B>();

shared_ptr calls B & A destructors while unique_ptr - only A's destructor. 

There should also be a difference in interface between shared_ptr & unique_ptr: in our case shared_ptr should take a reference to allocator to allocate service part(weak/shared counters, optional mutex & deleter). unique_ptr doesn't require that.

include/aos/common/tools/memory.hpp
```diff
@@ -122,66 +175,136 @@ class SmartPtr {
...
     }
 
     /**
      * Unique pointer move assignment for derived class.
      *
      * @param ptr unique pointer to assign from.
      */
-    template <typename P>
-    void operator=(UniquePtr<P>&& ptr)
+    template <typename P, typename Deleter2>
```
al1img, line 232: `Deleter2`  -> just `D`?

include/aos/common/tools/memory.hpp
MykolaSolyanko: > There is a subtle difference between deleter in unique & shared pointers, that makes it difficult to have a common implementation. Deleter for unique_ptr is stored on the stack, while shared_ptr's deleter stored in the heap(allocator's space in our case). SmartPtr implements a shared_ptr's approach, so I made a separate impl for unique_ptr.
> 
> There is also a difference in behaviour for the following example:
> 
> ```
> struct A{}; struct B: A {}
> unique_ptr<A> b = make_unique<B>();
> shared_ptr<A> b = make_shared<B>();
> ```
> 
> shared_ptr calls B & A destructors while unique_ptr - only A's destructor.
> 
> There should also be a difference in interface between shared_ptr & unique_ptr: in our case shared_ptr should take a reference to allocator to allocate service part(weak/shared counters, optional mutex & deleter). unique_ptr doesn't require that.

I did not get, in which compiler do this while unique_ptr - only A's destructor.? according to https://godbolt.org/, this call works correct  https://godbolt.org/z/qGvEn83E9
```
Base
Derived
```

MykolaSolyanko: > There is a subtle difference between deleter in unique & shared pointers, that makes it difficult to have a common implementation. Deleter for unique_ptr is stored on the stack, while shared_ptr's deleter stored in the heap(allocator's space in our case). SmartPtr implements a shared_ptr's approach, so I made a separate impl for unique_ptr.
> 
> There is also a difference in behaviour for the following example:
> 
> ```
> struct A{}; struct B: A {}
> unique_ptr<A> b = make_unique<B>();
> shared_ptr<A> b = make_shared<B>();
> ```
> 
> shared_ptr calls B & A destructors while unique_ptr - only A's destructor.
> 
> There should also be a difference in interface between shared_ptr & unique_ptr: in our case shared_ptr should take a reference to allocator to allocate service part(weak/shared counters, optional mutex & deleter). unique_ptr doesn't require that.

@mykola-kobets-epam  I did not get, why you decide unique_ptr - only A's destructor? https://godbolt.org/z/hrjev1v48 works well 
```
#include <iostream>
#include <string>
#include <memory>

struct Base {
   Base() {
      std::cout << "Base\n";
   }
};

struct Derived: Base {
   Derived() {
      std::cout << "Derived\n";
   }
};


int main() {
    std::unique_ptr<Base> ptr = std::make_unique<Derived>();
    return 0;
}

Base
Derived
```

include/aos/common/tools/memory.hpp
mykola-kobets-epam: changed.

include/aos/common/tools/memory.hpp
mykola-kobets-epam: SharedPtr refactoring should be done in a separate PR with possible implementation of weakptr

PR #116, opened 2024-04-26
add cppcheck action and fix issues

mlohvynenko:


include/aos/common/tools/memory.hpp
```diff
@@ -235,12 +235,14 @@ class SharedPtr : public SmartPtr<T> {
...
     }
 
+    // cppcheck-suppress-begin noExplicitConstructor
     /**
      * Creates shared pointer from another derived class shared pointer.
      *
      * @param ptr pointer to create from.
      */
     template <typename P>
+    // cppcheck-suppress noExplicitConstructor
```
mykola-kobets-epam, line 245: may be it would be better to make the constructor explicit?

mykola-kobets-epam (approved):

MykolaSolyanko (approved):

include/aos/common/tools/memory.hpp
mlohvynenko: llvm STL implementation doesn't mark template copy constructor as explicit.

https://github.com/llvm/llvm-project/blob/main/libcxx/include/__memory/shared_ptr.h#L571

Agreed with Oleksandr to follow the same  approach as llvm

include/aos/common/tools/time.hpp
```diff
@@ -239,10 +239,10 @@ class Time {
      */
     friend Log& operator<<(Log& log, const Time& obj)
     {
-        static constexpr auto cTimeStrLen = 32;
+        static constexpr auto cUTCTimeStrLen = 32;
```
al1img, line 242: Why it is renamed?

mlohvynenko: This constant shadows cTimeStrLen declared in the aos/comon/tools/time.hpp

al1img: Then cTimeStrLen should be used instead, no need to add new one. They look the same.

mlohvynenko: cTimeStrLen == 68
author of this code specified that string size will be 32.

So I see 2 options:
1. simply use cTimeStrLen global constant (68 bytes)
2. add a new constant (like AOS_CONFIG_TIME_UTS_STR_LEN 32) and use it

mlohvynenko: reused global constant.

al1img (approved):

PR #118, opened 2024-05-02
Mbedtls3.6 adaptation

mykola-kobets-epam:


include/aos/common/crypto/mbedtls/mbedtls_config.h
```diff
@@ -9,7 +9,19 @@
  */
```
MykolaSolyanko: I propose to merge this configuration with our config more accurately, because if we just copy past we can remove necessary in the future configuration from 3.6, for example MBEDTLS_PLATFORM_GMTIME_R_ALT etc

MykolaSolyanko (approved):

include/aos/common/crypto/mbedtls/mbedtls_config.h
mykola-kobets-epam: fixed

mlohvynenko (approved):

PR #125, opened 2024-05-31
NodeManager implementation

mykola-kobets-epam:


include/aos/iam/nodemanager/nodemanager.hpp
```diff
@@ -0,0 +1,252 @@
+/*
+ * Copyright (C) 2023 Renesas Electronics Corporation.
```
mlohvynenko, line 2:
```suggestion
 * Copyright (C) 2024 Renesas Electronics Corporation.
```

include/aos/iam/nodemanager/nodemanager.hpp
```diff
@@ -0,0 +1,252 @@
...
+#include "aos/iam/config.hpp"
+
+namespace aos::iam::nodemanager {
+
+/** @addtogroup IAM Node Manager
+ *  @{
+ */
+
+/**
+ * NodeInfo listener.
```
mlohvynenko, line 21:
```suggestion
 * NodeInfo listener interface.
```

include/aos/iam/nodemanager/nodemanager.hpp
```diff
@@ -0,0 +1,252 @@
...
+    const NodeInfo* GetNodeFromCache(const String& nodeId) const;
+
+    Error UpdateCache(const NodeInfo& nodeInfo);
+    Error UpdateCache(const String& id, NodeStatus status);
+
+    RetWithError<NodeInfo*> AddNodeInfoToCache();
+
+    void NotifyNodeInfoChange(const NodeInfo& nodeInfo);
+
+    static constexpr auto cNodeMaxNum = AOS_CONFIG_NODEMANAGER_NODE_MAX_NUM;
```
MykolaSolyanko, line 241: static const should be first in the private section according code style

src/iam/nodemanager/nodemanager.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+const NodeInfo* NodeManager::GetNodeFromCache(const String& nodeId) const
+{
+    return const_cast<NodeManager*>(this)->GetNodeFromCache(nodeId);
+}
+
+Error NodeManager::UpdateCache(const NodeInfo& nodeInfo)
+{
+    Error err     = ErrorEnum::eNone;
```
MykolaSolyanko, line 163: move err inside `if `

src/iam/nodemanager/nodemanager.cpp
```diff
@@ -0,0 +1,225 @@
...
+        *cachedInfo = nodeInfo;
+        NotifyNodeInfoChange(nodeInfo);
+    }
+
+    return ErrorEnum::eNone;
+}
+
+Error NodeManager::UpdateCache(const String& id, NodeStatus status)
+{
+    Error err     = ErrorEnum::eNone;
```
MykolaSolyanko, line 186: ditto

mlohvynenko (approved):

include/aos/iam/config.hpp
```diff
@@ -99,4 +99,11 @@
 #define AOS_CONFIG_PERMHANDLER_SECRET_LEN 42
 #endif
 
+/**
+ * Maximum number of nodes handled by node manager.
+ */
+#ifndef AOS_CONFIG_NODEMANAGER_NODE_MAX_NUM
+#define AOS_CONFIG_NODEMANAGER_NODE_MAX_NUM 3
```
al1img, line 106: Let's put 4.

include/aos/iam/nodemanager/nodemanager.hpp
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Destroys object instance.
+     */
+    virtual ~NodeManagerItf() = default;
+};
+
+/**
+ * Node info storage interface.
+ */
+class NodeInfoStorageItf {
```
al1img, line 109: Better to use CRUD operations for storage. We do not always have SQL as storage and set individual fields in storage may require more effort than just update it on client side. So more simple and usable API will look like:

```c++
AddNodeInfo
UpdateNodeInfo
GetNodeInfo
GetAllNodeIds
RemoveNodeInfo
```

src/iam/nodemanager/nodemanager.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+#include "aos/iam/nodemanager/nodemanager.hpp"
+#include "log.hpp"
+
+namespace aos::iam::nodemanager {
+
+/***********************************************************************************************************************
+ * Public
+ **********************************************************************************************************************/
+Error NodeManager::Init(NodeInfoStorageItf& storage)
```
al1img, line 16: Empty line after header.

src/iam/nodemanager/nodemanager.cpp
```diff
@@ -0,0 +1,227 @@
...
+    if (mNodeInfoListener) {
+        mNodeInfoListener->OnNodeRemoved(nodeId);
+    }
+
+    return ErrorEnum::eNone;
+}
+
+Error NodeManager::SubscribeNodeInfoChange(NodeInfoListenerItf& listener)
+{
+    for (const auto& nodeInfo : mNodeInfoCache) {
```
al1img, line 136: It is not expected that client will receive all updates on subscribe. Usually client will do something like that:
```c++
SubscribeNodeInfoChange
GetAllNodeInfos
```

src/iam/nodemanager/nodemanager.cpp
```diff
@@ -0,0 +1,227 @@
...
+        listener.OnNodeInfoChange(nodeInfo);
+    }
+
+    mNodeInfoListener = &listener;
+
+    return ErrorEnum::eNone;
+}
+
+NodeInfo* NodeManager::GetNodeFromCache(const String& nodeId)
```
al1img, line 145: Add private header section.

include/aos/iam/nodemanager/nodemanager.hpp
mykola-kobets-epam: making a crud interface makes it usage more difficult for client. I would better hide complexity inside an implementation behind a simpler interface.

src/iam/nodemanager/nodemanager.cpp
mykola-kobets-epam: done

src/iam/nodemanager/nodemanager.cpp
mykola-kobets-epam: done

src/iam/nodemanager/nodemanager.cpp
mykola-kobets-epam: Ok, So the question is whether it is required to send initial value when the listener subscribed to the manager.
I think the current approach is simpler for the manager & its clients, & as a result it is more correct. Unless there are some counterexamples from other frameworks, which might mean that there is a general approach & it differs from the implemented. So I suggest to leave it as is.

include/aos/iam/config.hpp
mykola-kobets-epam: done

include/aos/iam/nodemanager/nodemanager.hpp
mykola-kobets-epam: agreed to remove status(get/set) methods

src/iam/nodemanager/nodemanager.cpp
mykola-kobets-epam: Agreed not to send nodeInfo on subscription => the client should getallnodeinfo before subscription. Because there is no clear understanding when all the nodes are sent

tests/iam/nodemanager/storagemock.hpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (C) 2024 Renesas Electronics Corporation.
+ * Copyright (C) 2024 EPAM Systems, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef AOS_NODEINFO_STORAGE_HPP_
```
mlohvynenko, line 8:
```suggestion
#ifndef AOS_NODEINFO_STORAGE_MOCK_HPP_
```

mlohvynenko (approved):

tests/iam/nodemanager/storagemock.hpp
mykola-kobets-epam: fixed

mlohvynenko (approved):

al1img (approved):

PR #131, opened 2024-06-21
[cmake] Ignore stringop-truncation

mlohvynenko:


include/aos/common/tools/error.hpp
```diff
@@ -267,7 +267,10 @@ class Error {
     void CopyMessage(const char* msg)
     {
         if (msg != nullptr) {
+#pragma GCC diagnostic push
```
al1img, line 270: Disable this warning in main cmake file.

include/aos/common/tools/string.hpp
```diff
@@ -496,7 +496,7 @@ class String : public Array<char> {
         Clear();
 
         // cppcheck-suppress wrongPrintfScanfArgNum
-        auto ret = snprintf(Get(), MaxSize() + 1, format, args...);
```
al1img, line 499: Is there any warning here? Should be done in separate commit with proper commit message.

al1img (approved):

MykolaSolyanko (approved):

PR #134, opened 2024-06-29
[log] Update log module

MykolaSolyanko:


src/common/crypto/log.hpp
```diff
@@ -10,9 +10,11 @@
 
 #include "aos/common/tools/log.hpp"
 
-#define LOG_DBG() LOG_MODULE_DBG(aos::LogModuleEnum::eCommonCrypto)
-#define LOG_INF() LOG_MODULE_INF(aos::LogModuleEnum::eCommonCrypto)
-#define LOG_WRN() LOG_MODULE_WRN(aos::LogModuleEnum::eCommonCrypto)
-#define LOG_ERR() LOG_MODULE_ERR(aos::LogModuleEnum::eCommonCrypto)
+constexpr auto cLogModule = "crypto";
+
+#define LOG_DBG() LOG_MODULE_DBG(cLogModule)
```
al1img, line 15: Could we move these macros to `log.hpp` as well? Would be nice to have something like that:

```c++
#define LOG_MODULE "crypto"

#include "aos/common/tools/log.hpp"
```

Doesn't lool perfect and rather C style. Would be nice to avoid adding same macro in each module.

al1img (approved):

src/common/crypto/log.hpp
MykolaSolyanko: Updated

MykolaSolyanko: we could not move it in log.hpp because in some cases we include "log.hpp" as a separate include not like define logging, so I moved it to a separate header file

mlohvynenko (approved):

al1img (approved):

mykola-kobets-epam (approved):

PR #148, opened 2024-08-23
Trim newline at the end of the file

mykola-kobets-epam:


include/aos/common/tools/string.hpp
```diff
@@ -163,6 +163,22 @@ class String : public Array<char> {
         return Resize(from - begin());
     }
 
+    /**
+     * Removes trailing characters.
+     *
+     * @param chars characters to be removed.
+     */
+    void RTrim(const String& chars) {
```
al1img, line 171: Rename to `RightTrim`, add `LeftTrim` and `Trim` (both).

include/aos/common/tools/fs.hpp
```diff
@@ -421,7 +421,14 @@ class FS {
             return err;
         }
 
-        return text.Resize(buff.Size());
+        err = text.Resize(buff.Size());
```
al1img, line 424: Trim should be done  in caller. ReadFileToString should read what ever is in the file.

include/aos/common/tools/string.hpp
mykola-kobets-epam: fixed

include/aos/common/tools/fs.hpp
mykola-kobets-epam: removed PR

al1img (approved):

PR #151, opened 2024-09-03
FS unit tests

mykola-kobets-epam:


include/aos/common/tools/fs.hpp
```diff
@@ -364,33 +347,31 @@ class FS {
             return Error(errno);
         }
 
-        auto OnError = [fd]() {
-            auto err = errno;
-
+        auto OnError = [fd](const aos::Error& err) {
```
al1img, line 350: We have `DeferRelease` helper to release C resource. Use it here.

mykola-kobets-epam: fixed

include/aos/common/tools/fs.hpp
```diff
@@ -440,7 +418,7 @@ class FS {
             return err;
         }
 
-        auto fd = open(fileName.CStr(), O_CREAT | O_WRONLY, perm);
+        auto fd = open(fileName.CStr(), O_CREAT | O_WRONLY);
```
al1img, line 421: Why are permissions removed?

mykola-kobets-epam: Decided to do that because open is affected by umask for file creation. And it doesn't return an error if permissions of new file are different from requested.

al1img (approved):

mlohvynenko (approved):

PR #152, opened 2024-09-06
[common] Switch to mbedtls v3.6.1

mykola-kobets-epam:


mlohvynenko (approved):

MykolaSolyanko (approved):

src/common/crypto/mbedtls/drivers/psa_crypto.c
```diff
@@ -1,9176 +0,0 @@
-/*
```
al1img, line 1: Why this files is removed? It should be updated instead.

suppressions.txt
```diff
@@ -1,4 +1,3 @@
 useStlAlgorithm
 unusedFunction
-toomanyconfigs
```
al1img, line 3: Why?

src/common/crypto/mbedtls/drivers/psa_crypto.c
mykola-kobets-epam: done

suppressions.txt
mykola-kobets-epam: produced cpp-check error. might be due to missing psa_driver.c
https://github.com/aosedge/aos_core_lib_cpp/actions/runs/10739968945/job/29786976717
replaced by suppressing psa-driver dir entirely as it produces other warnings:

PR #155, opened 2024-09-12
Change pin-value to pin-source for pkcs11 url

mykola-kobets-epam:


src/common/cryptoutils.cpp
```diff
@@ -386,8 +386,21 @@ Error ParsePKCS11URL(
         return AOS_ERROR_WRAP(err);
     }
 
-    err = FindUrlParam(url, "pin-value", userPin);
-    if (!err.IsNone() && !err.Is(ErrorEnum::eNotFound)) {
+    StaticString<cFilePathLen> pinPath;
+
+    err = FindUrlParam(url, "pin-source", pinPath);
+    if (err.Is(ErrorEnum::eNotFound)) {
+        userPin.Clear();
```
mlohvynenko, line 393: why should user pin be cleared?
looks like it's empty at this stage.

mlohvynenko (approved):

src/common/cryptoutils.cpp
mykola-kobets-epam: it depends on the caller whether userPin empty or not. this check ensures, that it is indeed empty if pin-source is not provided.

al1img (approved):

PR #159, opened 2024-09-24
Renew certificate fix

mykola-kobets-epam:


include/aos/iam/certhandler.hpp
```diff
@@ -26,6 +24,55 @@ namespace certhandler {
  */
 constexpr auto cIAMCertModulesMaxCount = AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT;
 
+/**
+ * Max number of certificate subscriptions:
+ *
+ * 1 per client & server + cIAMCertModulesMaxCount for message proxy
+ */
+constexpr auto cIAMCertSubsMaxCount = cIAMCertModulesMaxCount * 2 + cIAMCertModulesMaxCount;
```
al1img, line 32: It should be private of CertHandler const. Also unclear why we need +cIAMCertModulesMaxCount

src/iam/certhandler/certhandler.cpp
```diff
@@ -140,21 +138,45 @@ Error CertHandler::GetCertificate(
...
     }
 
+    certReceiver.UpdateCert(certInfo);
+
     return ErrorEnum::eNone;
 }
 
+Error CertHandler::UnsubscribeCertReceiver(CertReceiverItf& certReceiver)
+{
+    LockGuard lock(mMutex);
```
al1img, line 167: LockGuard lock {mMutex};

include/aos/iam/certhandler.hpp
```diff
@@ -26,6 +24,55 @@ namespace certhandler {
...
+ * Certificate receiver interface.
+ */
+class CertReceiverItf {
+public:
+    /**
+     * Updates certificate.
+     *
+     * @param info certificate info.
+     */
+    void UpdateCert(const CertInfo& info)
```
al1img, line 44: It should not be part of interface. The client should have access only to required methods.

include/aos/iam/certhandler.hpp
mykola-kobets-epam: 1 per client & server + cIAMCertModulesMaxCount for message proxy

src/iam/certhandler/certhandler.cpp
mykola-kobets-epam: updated in a separate pr for all the files.

include/aos/iam/certhandler.hpp
mykola-kobets-epam: interface updated

src/iam/certhandler/certhandler.cpp
```diff
@@ -202,6 +241,23 @@ CertModule* CertHandler::FindModule(const String& certType) const
...
 
-} // namespace certhandler
-} // namespace iam
-} // namespace aos
+Error CertHandler::UpdateCerts(CertModule& certModule)
+{
+    CertInfo certInfo;
+
+    for (auto& subscription : mCertReceiverSubscriptions) {
+        auto err = certModule.GetCertificate(Array<uint8_t>(), Array<uint8_t>(), certInfo);
```
mlohvynenko, line 249: Didn't get why GetCertificate is called on each iteration, not before the loop

src/iam/certhandler/certhandler.cpp
```diff
@@ -155,28 +153,64 @@ Error CertHandler::GetCertificate(
...
+          });
+
+    return err;
+}
+
+Error CertHandler::CreateSelfSignedCert(const String& certType, const String& password)
+{
+    LockGuard lock(mMutex);
+
+    LOG_DBG() << "Create self signed cert: type = " << certType;
```
mlohvynenko, line 201: Remove spaces, so
`Create self signed cert: type='

src/iam/provisionmanager/provisionmanager.cpp
```diff
@@ -145,4 +145,18 @@ Error ProvisionManager::GetCert(
     return AOS_ERROR_WRAP(mCertHandler->GetCertificate(certType, issuer, serial, resCert));
 }
 
+Error ProvisionManager::SubscribeCertReceiver(const String& certType, certhandler::CertReceiverItf& certReceiver)
+{
+    LOG_DBG() << "Subscribe cert receiver: type = " << certType;
```
mlohvynenko, line 150: Remove spaces surrounding `=`

src/iam/certhandler/certhandler.cpp
mykola-kobets-epam: removed in a separate commit for the file

include/aos/iam/certhandler.hpp
```diff
@@ -26,6 +24,24 @@ namespace certhandler {
  */
 constexpr auto cIAMCertModulesMaxCount = AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT;
 
+/**
+ * Certificate receiver interface.
+ */
+class CertReceiverItf {
+public:
+    /**
+     * Process certificate updates.
```
al1img, line 33: Processes

include/aos/iam/certhandler.hpp
```diff
@@ -92,6 +108,23 @@ class CertHandlerItf {
...
         = 0;
 
+    /**
+     * Subscribes certificates receiver.
+     *
+     * @param certType certificate type.
+     * @param certReceiver certificate receiver.
+     * @returns Error.
+     */
+    virtual Error SubscribeCertReceiver(const String& certType, CertReceiverItf& certReceiver) = 0;
```
al1img, line 118: SubscribeCertChanged?

include/aos/iam/certhandler.hpp
```diff
@@ -221,16 +271,34 @@ class CertHandler : public CertHandlerItf, private NonCopyable {
     virtual ~CertHandler();
 
 private:
+    // (Two subscriptions for grpc client & server in iam/sm/cm/um) * modules count +
+    // (one subscription for grpc client in message proxy) * modules count
+    static constexpr auto cIAMCertSubsMaxCount = cIAMCertModulesMaxCount * 2 + cIAMCertModulesMaxCount;
```
al1img, line 276: Looks like in general we need only one subscription per storage. So cIAMCertModulesMaxCount should be enough. Also, we can't have this option as const. Add dedicated define to config.hpp with default value equals to modules max count.

mykola-kobets-epam: On the setup with main/secondary node  cIAMCertModulesMaxCount * 2 wasn't enough as I remember. And it looks like mp doesn't use IAMServer to get certificates, as well as IAMClient. But I guess, sm/cm/um clients - use. 
So I came up to the following formular: 
cIAMCertModulesMaxCount <for the servers> +  cIAMCertModulesMaxCount*cMaxNodesNum <for grpc clients> 

So we don't need additional configuration, the less configuration values we have - the better.

include/aos/iam/certhandler.hpp
mykola-kobets-epam: yes, it would be more uniform. changed

include/aos/iam/certhandler.hpp
mykola-kobets-epam: fixed

include/aos/iam/certhandler.hpp
al1img: I still do not understand why we need 3*cIAMCertModulesMaxCount subscriptions. Could you provide the example?

Please add dedicated configuration into config and set it to appropriate default value. Having config value doesn't mean configuring it all the time. But it make either to change this value in the feature.

mykola-kobets-epam: one for sm/um/cm grpc server in communication manager + one for iam grpc server in iammanager 
one for grpc client in sm/um/iam manager
And I was expected that message proxy also requested certificates using iam public interface. But it looks like it's not.
So we only require 2*cert-modules-max-count.

al1img: I see.  MP uses SM and IAM, client certificates and probably will use UM client certificates as well. We can't predict future client configuration. That's why config option is REQUIRED. And should be set to 3*AOS_CONFIG_CERTHANDLER_MODULES_MAX_COUNT.

mykola-kobets-epam: > Please add dedicated configuration into config and set it to appropriate default value. 
Added in a separate commit

al1img (approved):

mlohvynenko (approved):

src/iam/certhandler/certhandler.cpp
```diff
@@ -155,33 +153,69 @@ Error CertHandler::GetCertificate(
...
     auto* module = FindModule(certType);
     if (module == nullptr) {
         return AOS_ERROR_WRAP(ErrorEnum::eNotFound);
     }
 
-    auto err = module->CreateSelfSignedCert(password);
+    CertInfo certInfo;
+
+    auto err = module->GetCertificate(Array<uint8_t>(), Array<uint8_t>(), certInfo);
+    if (!err.IsNone()) {
```
mlohvynenko, line 170: Consider this change
```suggestion
    if (!err.IsNone() && !err.Is(ErrorEnum::eNotFound)) {
```

MykolaSolyanko (approved):

src/iam/certhandler/certhandler.cpp
mykola-kobets-epam: agreed not to do that

PR #164, opened 2024-10-07
[fs] Fix build for zephyr

mykola-kobets-epam:
Zephyr doesn't implement chmod command

### DIFF
--- a/include/aos/common/tools/fs.hpp
+++ b/include/aos/common/tools/fs.hpp
@@ -441,9 +441,12 @@ public:
             return Error(errno);
         }
 
+// Zephyr doesn't implement chmod
+#ifndef __ZEPHYR__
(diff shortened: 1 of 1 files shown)


mlohvynenko (approved):

include/aos/common/tools/fs.hpp
```diff
@@ -441,9 +441,12 @@ class FS {
             return Error(errno);
         }
 
+// Zephyr doesn't implement chmod
```
al1img, line 444: Why do we need this chmod at all?

mykola-kobets-epam: to fix permissions. open creates file with masked permissions, which can be different from requested.

al1img (approved):

PR #166, opened 2024-10-21
Adapt for submodules

MykolaSolyanko:


CMakeLists.txt
```diff
@@ -40,12 +40,17 @@ set(CMAKE_CXX_FLAGS "${common_flags} ${CMAKE_CXX_FLAGS}")
 set(CMAKE_C_FLAGS "${common_flags} ${CMAKE_C_FLAGS}")
 set(CMAKE_CXX_STANDARD 17)
 
+if(WITH_TEST)
```
al1img, line 43: It is not required as you can override it in the upper layer.

CMakeLists.txt
```diff
@@ -71,19 +76,11 @@ if(WITH_COVERAGE)
...
-# Includes
-# ######################################################################################################################
-
-include_directories(include)
-
 # ######################################################################################################################
 # Targets
 # ######################################################################################################################
 
-add_subdirectory(src/common)
```
al1img, line 84: * keep it as is;
* do not add cmake file to src: this lib consists of 4 main libs: common, iam, sm, and will have um, cm in the feature. Basically, client of this lib requires only one part of the lib (+common). For example: iam_cpp will include common and iam only, sm_cpp will include common and sm etc.

include/aos/common/tools/error.hpp
```diff
@@ -9,6 +9,7 @@
 #define AOS_ERROR_HPP_
 
 #include <errno.h>
+#include <stdio.h>
```
al1img, line 12: Move these changes to the separate commit.

CMakeLists.txt
MykolaSolyanko: Unfortunately, this is necessary because these changes will not be overridden. Using the main CMake file might be possible when running tests separately, meaning we would need a separate build. Therefore, these will be distinct changes for CMake and the build, different from the main configuration.

CMakeLists.txt
MykolaSolyanko: Done

include/aos/common/tools/error.hpp
MykolaSolyanko: Done

CMakeLists.txt
al1img: We should find another solution. Let's discuss.

PR #167, opened 2024-11-07
Fix services restart

al1img:


include/aos/common/tools/list.hpp
```diff
@@ -0,0 +1,616 @@
...
+        if (!node) {
+            return nullptr;
+        }
+
+        new (&node->mValue) T(args...);
+
+        return node;
+    }
+
+    void InsertNode(Node& posNode, Node& newNode)
```
mykola-kobets-epam, line 258: insert in the middle might fail

include/aos/common/tools/list.hpp
```diff
@@ -0,0 +1,616 @@
...
+         */
+        ConstIterator() = default;
+
+        /**
+         *  Creates const iterator.
+         *
+         * @param node list node.
+         */
+        explicit ConstIterator(const Node* node)
+            : mCurrentNode(const_cast<Node*>(node))
```
mykola-kobets-epam, line 145: mCurrentNode can be a pointer to const to avoid const-cast

include/aos/common/tools/list.hpp
```diff
@@ -0,0 +1,616 @@
...
+            return it;
+        }
+
+        /**
+         * Checks if iterators are not equal.
+         *
+         * @param other iterator to compare with.
+         * @return bool.
+         */
+        bool operator!=(const ConstIterator& other) const { return mCurrentNode != other.mCurrentNode; }
```
mykola-kobets-epam, line 214: operator ==  should be added for consistency

include/aos/common/tools/list.hpp
```diff
@@ -0,0 +1,616 @@
...
+        bool  mAllocated = false;
+        T     mValue;
+        Node* mNext = nullptr;
+        Node* mPrev = nullptr;
+    };
+
+    /**
+     * List iterator.
+     */
+    class Iterator {
```
mykola-kobets-epam, line 39: It seems like the difference between ConstIterator & Iterator is in dereference operator.
A single template should be enough to cover both cases

include/aos/common/tools/list.hpp
```diff
@@ -0,0 +1,616 @@
...
+    {
+        mTerminalNode.mNext = &mTerminalNode;
+        mTerminalNode.mPrev = &mTerminalNode;
+    }
+
+    ListImpl(const ListImpl& list) = default;
+
+    virtual Node* AllocateNode() = 0;
+
+    Node* AllocateNode(const T& item)
```
mykola-kobets-epam, line 233: I guess AllocateNode(Args&&... args) should be enough to cover both cases.
Also it might better rename the method to CreateNode.

include/aos/common/tools/algorithm.hpp
```diff
@@ -0,0 +1,295 @@
...
+namespace aos {
+
+/**
+ * Algorithm interface.
+ * @tparam T value type.
+ * @tparam I iterator type.
+ * @tparam CI const iterator type.
+ */
+template <typename T, typename I, typename CI>
+class AlgorithmItf {
```
mykola-kobets-epam, line 22: I suppose it's better not to create a class for container utils, but keep them as  separate functions.

include/aos/common/tools/allocator.hpp
```diff
@@ -11,8 +11,8 @@
 #include <assert.h>
```
mykola-kobets-epam, line 11: I think better create a Set class for allocations

include/aos/common/tools/algorithm.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Finds const element in container.
+     *
+     * @param item element to find.
+     * @return RetWithError<CI>.
+     */
+    RetWithError<CI> Find(const T& item) const
+    {
+        for (auto it = begin(); it != end(); it++) {
```
mlohvynenko, line 176: Consider using pre-increment for all these loop iterators.

include/aos/common/tools/list.hpp
al1img: redesigned

include/aos/common/tools/list.hpp
al1img: done

include/aos/common/tools/list.hpp
al1img: Redesigned.

include/aos/common/tools/list.hpp
al1img: done

include/aos/common/tools/algorithm.hpp
al1img: Why? What would be the benefit?

include/aos/common/tools/algorithm.hpp
al1img: done

include/aos/common/tools/allocator.hpp
al1img: In general yes, but from current allocator implementation view, there is no advantages using set. Also, list has potentially more use cases in the future.

include/aos/common/tools/algorithm.hpp
mykola-kobets-epam: In case of a base class approach, it would better utilize CRTP, but with common type aliases added to parent container classes for convenience:  

class Array<T> : public BaseContainerItf< Array<T> > {
    using Val = T;
    using ConstIter = const T*;
    using Iter = T*;
};

class BaseContainerItf< ParentCont > {
   ParentCont* parent() const { return static_cast<const ParentCont*>(this); }
   bool IsEmpty() const { return parent()->Size() == 0; }
   const typename ParentCont::Val& Front() const {return *parent->begin(); }
}

For approach with external functions. Target container should be added as the first argument. eg.
template <class C>
typename const C::Val& Front(const C& cont) { return *cont.begin(); }

mykola-kobets-epam (approved):

include/aos/common/tools/algorithm.hpp
al1img:
```c++
typename const C::Val& Front(const C& cont) { return *cont.begin(); }

Front(mList) 
```
Looks weird. I prefer `mList.Front()`.

mlohvynenko (approved):

PR #171, opened 2024-11-25
Fix slot numbering

MykolaSolyanko:
Previously, when a slot was removed, and a new slot was added, the new slot would receive a number already present in the container. For example, if there were two slots with numbers 0 and 1, and slot 0 was deleted, adding a new slot would incorrectly assign it the number 1, which was already in use.

### DIFF
--- a/include/aos/sm/servicemanager.hpp
+++ b/include/aos/sm/servicemanager.hpp
@@ -82,7 +82,8 @@ struct ServiceData {
     bool operator==(const ServiceData& data) const
     {
         return mServiceID == data.mServiceID && mProviderID == data.mProviderID && mVersion == data.mVersion
-            && mImagePath == data.mImagePath && mCached == data.mCached && mSize == data.mSize && mGID == data.mGID;
+            && mImagePath == data.mImagePath && data.mManifestDigest == mManifestDigest && mCached == data.mCached
+            && mSize == data.mSize && mGID == data.mGID;
     }
 
     /**
--- a/src/common/crypto/mbedtls/driverwrapper.cpp
+++ b/src/common/crypto/mbedtls/driverwrapper.cpp
@@ -27,26 +27,28 @@
  * Key description.
  */
 struct KeyDescription {
+    KeyDescription() = default;
+
     /**
      * Key ID.
      */
-    psa_key_id_t mKeyID;
+    psa_key_id_t mKeyID {};
     /**
      * Key lifetime.
      */
-    psa_key_lifetime_t mLifetime;
-    /**
-     * Slot number.
-     */
-    psa_drv_slot_number_t mSlotNumber;
+    psa_key_lifetime_t mLifetime {};
     /**
      * Hash algorithm.
      */
-    aos::crypto::HashEnum mHashAlg;
+    aos::crypto::HashEnum mHashAlg {};
+    /**
+     * Free key flag.
+     */
+    bool mAllocated {};
     /**
      * Private key.
      */
-    const aos::crypto::PrivateKeyItf* mPrivKey;
+    const aos::crypto::PrivateKeyItf* mPrivKey {};
 };
 
 /***********************************************************************************************************************
@@ -58,8 +60,9 @@ static constexpr psa_algorithm_t sPSAAlgs[]
 static constexpr mbedtls_md_type_t sMDTypes[]
     = {MBEDTLS_MD_SHA1, MBEDTLS_MD_SHA224, MBEDTLS_MD_SHA256, MBEDTLS_MD_SHA384, MBEDTLS_MD_SHA512};
 
-static aos::StaticArray<KeyDescription, MBEDTLS_PSA_KEY_SLOT_COUNT> sBuiltinKeys;
-static aos::Mutex                                                   sMutex;
+static aos::StaticArray<KeyDescription, MBEDTLS_PSA_KEY_SLOT_COUNT> sBuiltinKeys(MBEDTLS_PSA_KEY_SLOT_COUNT);
+
+static aos::Mutex sMutex;
 
 static int ExportRSAPublicKeyToDER(
     const aos::crypto::RSAPublicKey& rsaKey, uint8_t* data, size_t dataSize, size_t* dataLength)
@@ -222,6 +225,23 @@ static int ExportECPublicKeyToDER(
     return 0;
 }
 
+static bool IsKeyInBuiltinList(psa_key_id_t keyID)
+{
+    auto ret = sBuiltinKeys.FindIf([&](const KeyDescription& key) { return key.mKeyID == keyID; });
+
+    return ret.mError.IsNone();
+}
+
+static aos::RetWithError<KeyDescription*> FindFreeSlot()
+{
+    auto ret = sBuiltinKeys.FindIf([](const KeyDescription& key) { return !key.mAllocated; });
+    if (!ret.mError.IsNone()) {
+        return aos::RetWithError<KeyDescription*>(nullptr, aos::ErrorEnum::eNotFound);
+    }
+
+    return aos::RetWithError<KeyDescription*>(ret.mValue, aos::ErrorEnum::eNone);
+}
+
 aos::crypto::HashEnum GetRSASHAAlgorithm(size_t modulusBitlen)
 {
     if (modulusBitlen < 2048) {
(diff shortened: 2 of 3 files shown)


mlohvynenko (approved):

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -300,12 +300,18 @@ aos::RetWithError<KeyInfo> AosPsaAddKey(const aos::crypto::PrivateKeyItf& privKe
...
+        psa_drv_slot_number_t maxSlotNumber = 0;
+
         for (auto& key : sBuiltinKeys) {
             if (key.mKeyID == keyID) {
                 found = true;
 
                 break;
             }
+
+            if (key.mSlotNumber > maxSlotNumber) {
```
al1img, line 312: It means that the slot number is dynamic. If you remove one key, add another, and so on, the slot number will always increase. I strongly feel that slot number should be index in `sBuiltinKeys` array. E.g. `mSlotNumber` should not be part of `KeyDescription` but rather should be some flag as `mIntialized` for example. And when key is deleted this flag is reset when key added this flag is set and corresponding fields are updated.

mykola-kobets-epam (approved):

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -27,26 +27,28 @@
...
+     * Hash algorithm.
      */
-    psa_drv_slot_number_t mSlotNumber;
+    aos::crypto::HashEnum mHashAlg {};
     /**
-     * Hash algorithm.
+     * Free key flag.
      */
-    aos::crypto::HashEnum mHashAlg;
+    bool mFree {true};
```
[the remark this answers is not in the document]
MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -58,8 +60,8 @@ static constexpr psa_algorithm_t sPSAAlgs[]
 static constexpr mbedtls_md_type_t sMDTypes[]
     = {MBEDTLS_MD_SHA1, MBEDTLS_MD_SHA224, MBEDTLS_MD_SHA256, MBEDTLS_MD_SHA384, MBEDTLS_MD_SHA512};
 
-static aos::StaticArray<KeyDescription, MBEDTLS_PSA_KEY_SLOT_COUNT> sBuiltinKeys;
-static aos::Mutex                                                   sMutex;
+static KeyDescription sBuiltinKeys[MBEDTLS_PSA_KEY_SLOT_COUNT] {};
```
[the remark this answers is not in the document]
MykolaSolyanko: Done

include/aos/sm/servicemanager.hpp
```diff
@@ -82,7 +82,8 @@ struct ServiceData {
     bool operator==(const ServiceData& data) const
     {
         return mServiceID == data.mServiceID && mProviderID == data.mProviderID && mVersion == data.mVersion
-            && mImagePath == data.mImagePath && mCached == data.mCached && mSize == data.mSize && mGID == data.mGID;
+            && mImagePath == data.mImagePath && data.mManifestDigest == mManifestDigest && mTimestamp == data.mTimestamp
```
mlohvynenko, line 85: Not sure mTimestamp comparation is a good idea
Timestamp is usually the time object was created.
Only DB has the proper date.

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -421,45 +448,50 @@ psa_status_t aos_get_builtin_key(psa_drv_slot_number_t slotNumber, psa_key_attri
...
-            return keyBufferSize != 0 ? PSA_SUCCESS : PSA_ERROR_BUFFER_TOO_SMALL;
-        }
+        return PSA_ERROR_NOT_SUPPORTED;
     }
 
+    psa_set_key_id(attributes, sBuiltinKeys[slotNumber].mKeyID);
+    psa_set_key_lifetime(attributes, sBuiltinKeys[slotNumber].mLifetime);
+    psa_set_key_usage_flags(attributes, PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_VERIFY_HASH);
+
+    return keyBufferSize != 0 ? PSA_SUCCESS : PSA_ERROR_BUFFER_TOO_SMALL;
```
mlohvynenko, line 493: looks like method's code below return can be removed

include/aos/sm/servicemanager.hpp
MykolaSolyanko: we use this comparation in DB, and also we should be sure that object is equal. @al1img  need your opinion

src/common/crypto/mbedtls/driverwrapper.cpp
MykolaSolyanko: Done

include/aos/sm/servicemanager.hpp
MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -222,6 +225,24 @@ static int ExportECPublicKeyToDER(
...
+{
+    auto ret = sBuiltinKeys.FindIf([&](const KeyDescription& key) { return key.mKeyID == keyID; });
+
+    return ret.mError.IsNone();
+}
+
+static aos::RetWithError<size_t> FindFreeSlot()
+{
+    auto ret = sBuiltinKeys.FindIf([](const KeyDescription& key) { return !key.mAllocated; });
+
```
al1img, line 238: Remove empty line

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -222,6 +225,24 @@ static int ExportECPublicKeyToDER(
...
+    if (ret.mError.IsNone()) {
```
al1img, line 239: Handle error first

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -222,6 +225,24 @@ static int ExportECPublicKeyToDER(
...
 }
 
+static bool IsKeyInBuiltinList(psa_key_id_t keyID)
+{
+    auto ret = sBuiltinKeys.FindIf([&](const KeyDescription& key) { return key.mKeyID == keyID; });
+
+    return ret.mError.IsNone();
+}
+
+static aos::RetWithError<size_t> FindFreeSlot()
```
al1img, line 235: Why not return iterator instead of calculating index?

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -298,69 +319,71 @@ aos::RetWithError<KeyInfo> AosPsaAddKey(const aos::crypto::PrivateKeyItf& privKe
...
+        }
 
-        for (auto& key : sBuiltinKeys) {
-            if (key.mKeyID == keyID) {
-                found = true;
+        auto [slotNumber, err] = FindFreeSlot();
 
-                break;
+        if (!err.IsNone()) {
+            if (err = sBuiltinKeys.PushBack(KeyDescription()); !err.IsNone()) {
```
al1img, line 329: We can initialize array with max size. But it requires to add this type of constructor to StaticArray.

MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
MykolaSolyanko: Done

src/common/crypto/mbedtls/driverwrapper.cpp
```diff
@@ -298,69 +319,71 @@ aos::RetWithError<KeyInfo> AosPsaAddKey(const aos::crypto::PrivateKeyItf& privKe
...
+            continue;
+        }
 
-        for (auto& key : sBuiltinKeys) {
-            if (key.mKeyID == keyID) {
-                found = true;
+        auto [slotNumber, err] = FindFreeSlot();
 
-                break;
+        if (!err.IsNone()) {
```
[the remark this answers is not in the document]
MykolaSolyanko: Done

al1img (approved):

PR #174, opened 2024-11-27
[array] Prevent undefined behavior in loop

MykolaSolyanko:


mlohvynenko (approved):

include/aos/common/tools/array.hpp
```diff
@@ -326,7 +326,7 @@ class Array : public AlgorithmItf<T, T*, const T*> {
      */
     Error Insert(T* pos, const T* from, const T* till)
     {
-        auto size = till - from;
```
mlohvynenko, line 329: consider adding assert from > till

al1img, line 329: Unclear, why does undefined behavior occur in the current implementation?

include/aos/common/tools/array.hpp
```diff
@@ -326,7 +326,7 @@ class Array : public AlgorithmItf<T, T*, const T*> {
...
+        const size_t size = till - from;
```
al1img, line 329: could it be `const auto`?

include/aos/common/tools/array.hpp
MykolaSolyanko: this error message from yocto build

iteration 66 invokes undefined behavior [-Werror=aggressive-loop-optimizations]
|   343 |         for (auto i = 0; i < size; i++) {
|       |         ^~~
| /home/mykola-solianko/epam/new_aos_vm_version/meta-aos-vm/yocto/build-main/tmp/work/core2-64-aosvm-linux/aos-messageproxy/git/git/external/aos_core_common_cpp/external/aos_core_lib_cpp/include/aos/common/tools/array.hpp:343:28: note: within this loop
|   343 |         for (auto i = 0; i < size; i++) {
|       |                          ^~~~
| cc1plus: all warnings being treated as errors


so this variable should typed because when you do diff in pointer it one type but `auto i` is another type

al1img (approved):

PR #187, opened 2024-12-20
[networkmanager] Use enum for SetTrafficPeriod

MykolaSolyanko:


include/aos/sm/networkmanager.hpp
```diff
@@ -183,6 +183,8 @@ class StorageItf {
     virtual ~StorageItf() = default;
 };
 
+enum class TrafficPeriod { MinutePeriod, HourPeriod, DayPeriod, MonthPeriod, YearPeriod };
```
al1img, line 186: Use Aos enum

MykolaSolyanko: Done

al1img (approved):

PR #200, opened 2024-12-25
 Small fixes

al1img:


tests/sm/networkmanager/networkmanager_test.cpp
```diff
@@ -80,13 +76,13 @@ class NetworkManagerTest : public Test {
...
-    TrafficMonitorMock                           mTrafficMonitor;
-    NetworkManager                               mNetManager;
-    std::unique_ptr<NamespaceManagerMock>        mNetns;
-    std::unique_ptr<NetworkInterfaceManagerMock> mNetIf;
-    aos::StaticString<aos::cFilePathLen>         mWorkingDir;
+    StorageMock                          mStorage;
+    CNIMock                              mCNI;
+    TrafficMonitorMock                   mTrafficMonitor;
+    NetworkManager                       mNetManager;
+    NamespaceManagerMock                 mNetns;
```
MykolaSolyanko, line 83: why you delete StrictMock ? this good to monitor if all mock called as expected, in case if we forgot to call mock test will be failed, and I did it to be sure that all mock function will be called in case of error add instance to network

include/aos/common/spaceallocator/spaceallocator.hpp
```diff
@@ -114,6 +114,107 @@ class SpaceAllocatorItf {
     virtual ~SpaceAllocatorItf() = default;
 };
 
+/**
+ * Space instance.
+ */
+class Space : public SpaceItf {
```
mlohvynenko, line 120: Let's create Space with the given space, return not 0, but some meaningful data.
Also, let's implement resize logic (simply change underlaying data).

With this changes, such a Space object will be useful for Layer/Service manager, `cause image::ImageHandler returns space object, from which Size is taken to store it in DB.

tests/include/mocks/certloadermock.hpp
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (C) 2024 Renesas Electronics Corporation.
```
mlohvynenko, line 2: Shouldn't we remove renesas copyright from new files?

mykola-kobets-epam (approved):

mlohvynenko (approved):

MykolaSolyanko (approved):

PR #207, opened 2024-12-30
CNI wrapper

MykolaSuperman:


mykola-kobets-epam (approved):

mlohvynenko (approved):

include/aos/sm/cni.hpp
```diff
@@ -277,6 +277,15 @@ class CNIItf {
...
     virtual Error ValidateNetworkList(const NetworkConfigList& net) = 0;
+
+    /**
+     * Returns network list from cache.
+     *
+     * @param[out] net Network list.
+     * @param[out] rt Runtime configuration.
+     * @return Error.
+     */
+    virtual Error GetNetworkListCachedConfig(NetworkConfigList& net, RuntimeConf& rt) = 0;
```
al1img, line 288: Shouldn't net parameter be input and const?

MykolaSolyanko: No, they shouldn’t be like that because they are filled inside this method.

mlohvynenko (approved):

al1img (approved):

PR #231, opened 2025-01-17
[all] Return iterator in find methods

mlohvynenko:


mykola-kobets-epam (approved):

src/sm/launcher/launcher.cpp
```diff
@@ -727,6 +728,15 @@ void Launcher::StopInstances(const Array<InstanceData>& instances)
     if (auto err = mLaunchPool.Wait(); !err.IsNone()) {
         LOG_ERR() << "Launch pool wait error: err=" << err;
     }
+
+    {
+        UniqueLock lock {mMutex};
```
al1img, line 728: LockGuard

al1img (approved):

PR #233, opened 2025-01-20
[common] Move Time::ToString to crypto.hpp

mykola-kobets-epam:


include/aos/common/crypto/crypto.hpp
```diff
@@ -467,6 +467,14 @@ struct Extension {
     bool operator!=(const Extension& extension) const { return !operator==(extension); }
 };
 
+/**
+ * Converts input time to ASN1 GeneralizedTime string.
+ *
+ * @param time time.
+ * @return RetWithError<StaticString<cTimeStrLen>>
+ */
+RetWithError<StaticString<cTimeStrLen>> ConvertTimeToString(const Time& time);
```
al1img, line 476: Probably, better to move to crypto/utils.hpp?

`ConvertTimeToASN1Str` ?

mykola-kobets-epam: changed, but still have doubts whether asn1 infix required taking into account that the function included into asn1 namespace

src/common/crypto/crypto.cpp
```diff
@@ -0,0 +1,36 @@
+
+
```
mlohvynenko, line 2: Remove these 2 empty lines at the beginning.

mlohvynenko (approved):

al1img (approved):

mlohvynenko (approved):

src/common/crypto/crypto.cpp
mykola-kobets-epam: done

PR #250, opened 2025-01-30
[common,tools] Implement duration as a class

mlohvynenko:
Duration object log examples:

```
[ RUN      ] TimeTest.DurationToString
debug | default | Duration: -P6D
debug | default | Duration: P1W
debug | default | Duration: P1M
debug | default | Duration: P1Y
debug | default | Duration: P1Y1M1W1DT1H
debug | default | Duration: PT0S
debug | default | Duration: PT0.000000001S
debug | default | Duration: PT1M1S
debug | default | Duration: PT1M0.000032000S
```

### DIFF
--- a/include/aos/common/tools/time.hpp
+++ b/include/aos/common/tools/time.hpp
@@ -18,14 +18,268 @@
 namespace aos {
 
 /**
- * Base type for a time duration in nanoseconds. Can also be negative to set a point back in time.
- */
-using Duration = int64_t;
-
-/**
  * Size of a time in string representation.
  */
 const auto cTimeStrLen = AOS_CONFIG_TIME_STR_LEN;
+
+/**
+ * Base type for a time duration in nanoseconds. Can also be negative to set a point back in time.
+ */
+class Duration {
+public:
+    /**
+     * Default constructor.
+     */
+    constexpr Duration() = default;
+
+    // cppcheck-suppress noExplicitConstructor
+    /**
+     * Constructs object instance.
+     *
+     * @param duration duration in nanoseconds.
+     */
+    constexpr Duration(int64_t duration)
+        : mDuration(duration)
+    {
+    }
+
+    /**
+     * Returns duration in nanoseconds.
+     *
+     * @result int64_t.
+     */
+    int64_t Nanoseconds() const { return mDuration; }
+
+    /**
+     * Returns duration in microseconds.
+     *
+     * @result int64_t.
+     */
+    int64_t Microseconds() const { return Nanoseconds() / 1000; }
+
+    /**
+     * Returns duration in milliseconds.
+     *
+     * @result int64_t.
+     */
+    int64_t Milliseconds() const { return Microseconds() / 1000; }
+
+    /**
+     * Returns duration in seconds.
+     *
+     * @result float.
+     */
+    float Seconds() const { return static_cast<float>(Nanoseconds()) / 1000000000.f; }
+
+    /**
+     * Returns duration in minutes.
+     *
+     * @result float.
+     */
+    float Minutes() const { return Seconds() / 60.f; }
+
+    /**
+     * Returns duration in hours.
+     *
+     * @result float.
+     */
+    float Hours() const { return Minutes() / 60.f; }
+
+    /**
+     * Adds duration to the current object.
+     *
+     * @param duration duration.
+     * @result Duration&.
+     */
+    constexpr Duration& operator+=(const Duration& duration)
+    {
+        mDuration += duration.mDuration;
+
+        return *this;
+    }
+
+    /**
+     * Subtracts duration from the current object.
+     *
+     * @param duration duration.
+     * @result Duration&.
+     */
+    constexpr Duration& operator-=(const Duration& duration)
+    {
+        mDuration -= duration.mDuration;
+
+        return *this;
+    }
+
+    /**
+     * Multiplies duration.
+     *
+     * @param multiplier multiplier.
+     * @result Duration&.
+     */
+    constexpr Duration& operator*=(int64_t multiplier)
+    {
+        mDuration *= multiplier;
+
+        return *this;
+    }
+
+    /**
+     * Multiplies duration.
+     *
+     * @param multiplier multiplier.
+     * @param duration duration.
+     * @result Duration.
+     */
+    friend constexpr Duration operator*(int64_t multiplier, Duration duration) { return duration * multiplier; }
+
+    /**
+     * Divides duration.
+     *
+     * @param divider divider.
+     * @result Duration&.
+     */
+    constexpr Duration& operator/=(int64_t divider)
+    {
+        assert(divider != 0);
+
+        mDuration /= divider;
+
+        return *this;
+    }
+
+    /**
+     * Divides duration.
+     *
+     * @param divider divider.
+     * @param duration duration.
+     * @result Duration.
+     */
+    friend constexpr Duration operator/(int64_t divider, Duration duration)
+    {
+        assert(divider != 0);
+
+        return duration / divider;
+    }
+
+    /**
+     * Changes sign of the duration.
+     *
+     * @result Duration.
+     */
+    constexpr Duration operator-() const { return -mDuration; }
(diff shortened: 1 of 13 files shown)


include/aos/common/tools/time.hpp
```diff
@@ -101,14 +250,14 @@ class Time {
     {
         auto time = mTime;
 
-        int64_t nsec = time.tv_nsec + duration;
+        int64_t nsec = time.tv_nsec + duration.Count();
```
MykolaSolyanko, line 253: maybe an addition to `Count` method could also be better to add the convert `operator` to `int_64` or similar than in some cases not needed to call explicitly `Count` method

MykolaSuperman (approved):

include/aos/common/tools/time.hpp
mlohvynenko: I'd prefer to call Count explicitly cause it decreases  type safety.
As a pros:  no need to add arithmetic operations overloads for the Duration class.

include/aos/common/tools/time.hpp
```diff
@@ -17,15 +17,161 @@
...
 /**
  * Base type for a time duration in nanoseconds. Can also be negative to set a point back in time.
  */
-using Duration = int64_t;
+class Duration {
+public:
+    /**
+     * Default constructor.
+     */
+    constexpr Duration() = default;
```
al1img, line 33: What is benefit to use constexpr here?

include/aos/common/tools/time.hpp
```diff
@@ -17,15 +17,161 @@
...
+        : mDuration(duration)
+    {
+    }
+
+    /**
+     * Returns duration in nanoseconds.
+     *
+     * @result int64_t.
+     */
+    constexpr int64_t Count() const { return mDuration; }
```
al1img, line 51: It is not obvious that Count is in nanoseconds. So I propose to have such method as:

```c++
uint64_t Nanoseconds() - return duration in nanoseconds
uint64_t Microseconds() - return duration in microseconds
uint64_t Milliseconds() - return duration in milliseconds
float Seconds() - return duration in float seconds
float Minutes() - return duration in float minutes
float Hours() - return duration in float hours
```

include/aos/common/tools/time.hpp
```diff
@@ -17,15 +17,161 @@
...
 /**
- * Size of a time in string representation.
+ * Adds two durations and returns result as a new object.
+ *
+ * @param lhs left hand side duration.
+ * @param rhs right hand side duration.
+ * @result Duration.
  */
-const auto cTimeStrLen = AOS_CONFIG_TIME_STR_LEN;
+constexpr Duration operator+(const Duration& lhs, const Duration& rhs)
```
al1img, line 135: Could all these operators be places inside Duration class?

include/aos/common/tools/time.hpp
```diff
@@ -17,15 +17,161 @@
...
+}
+
+/**
+ * Multiplies two durations and returns result as a new object.
+ *
+ * @param lhs left hand side duration.
+ * @param rhs right hand side duration.
+ * @result Duration.
+ */
+constexpr Duration operator*(const Duration& lhs, const Duration& rhs)
```
al1img, line 159: Duration * Duration doesn't make sense. We should have Duration * int instead.

include/aos/common/tools/time.hpp
```diff
@@ -17,15 +17,161 @@
...
+}
+
+/**
+ * Divides two durations and returns result as a new object.
+ *
+ * @param lhs left hand side duration.
+ * @param rhs right hand side duration.
+ * @result Duration.
+ */
+constexpr Duration operator/(const Duration& lhs, const Duration& rhs)
```
al1img, line 171: Ditto

include/aos/common/tools/time.hpp
mlohvynenko: It's required for constructor that accepts int64_t, so it  can be called in a constexpr context.
Like this one:
`static constexpr Duration cNanoseconds  = 1;`

As for the default one, I've decided to use the same style. Also, I see no disadvantages of using constexpr if it's possible.

include/aos/common/tools/time.hpp
```diff
@@ -18,14 +18,245 @@
...
+     * @result int64_t.
+     */
+    int64_t Milliseconds() const { return Microseconds() / 1000; }
+
+    /**
+     * Returns duration in seconds.
+     *
+     * @result float.
+     */
+    float Seconds() const { return Milliseconds() / 1000; }
```
al1img, line 72: Should be `Nanoseconds() / 1000000000` also add .0 or f: `Nanoseconds() / 1000000000.0`. Otherwise, it perform int division and then cast to float if I'm not mistaken.

al1img (approved):

include/aos/common/monitoring/resourcemonitor.hpp
```diff
@@ -20,8 +20,8 @@ namespace aos::monitoring {
  * Resource monitor config.
  */
 struct Config {
-    Duration mPollPeriod    = AOS_CONFIG_MONITORING_POLL_PERIOD_SEC * Time::cSeconds;
-    Duration mAverageWindow = AOS_CONFIG_MONITORING_AVERAGE_WINDOW_SEC * Time::cSeconds;
+    Duration mPollPeriod    = Time::cSeconds * AOS_CONFIG_MONITORING_POLL_PERIOD_SEC;
```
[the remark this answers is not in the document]
mlohvynenko: added friend operators.

include/aos/common/tools/time.hpp
```diff
@@ -41,7 +295,10 @@ class Time {
     static constexpr Duration cSeconds      = 1000 * cMilliseconds;
     static constexpr Duration cMinutes      = 60 * cSeconds;
     static constexpr Duration cHours        = 60 * cMinutes;
```
mykola-kobets-epam, line 297: I suggest putting all duration consts into a singular form, same as new ones(Day,Week,..

include/aos/common/tools/timer.hpp
```diff
@@ -71,12 +71,12 @@ class Timer {
         sev.sigev_value.sival_ptr = this;
         sev.sigev_notify_function = TimerFunction;
 
-        its.it_value.tv_sec  = interval / Time::cSeconds;
-        its.it_value.tv_nsec = interval % Time::cSeconds;
+        its.it_value.tv_sec  = interval.Nanoseconds() / Time::cSeconds.Nanoseconds();
```
mykola-kobets-epam, line 74: Im fine with that, but can be changed to (interval / cSecond).Nanoseconds().

include/aos/sm/layermanager.hpp
```diff
@@ -186,7 +186,7 @@ struct Config {
     StaticString<cFilePathLen> mLayersDir;
     StaticString<cFilePathLen> mDownloadDir;
     Duration                   mTTL;
-    Duration                   mRemoveOutdatedPeriod = 24 * Time::cHours;
+    Duration                   mRemoveOutdatedPeriod = Time::cHours * 24;
```
mykola-kobets-epam, line 189: Time::cDay

include/aos/sm/servicemanager.hpp
```diff
@@ -250,7 +250,7 @@ struct Config {
     StaticString<cFilePathLen> mServicesDir;
     StaticString<cFilePathLen> mDownloadDir;
     Duration                   mTTL;
-    Duration                   mRemoveOutdatedPeriod = 24 * Time::cHours;
+    Duration                   mRemoveOutdatedPeriod = Time::cHours * 24;
```
mykola-kobets-epam, line 253: Time::cDay

mykola-kobets-epam (approved):

include/aos/common/tools/time.hpp
mlohvynenko: I'd keep as it is or if you insist It can be done as a separate task. (to reduce code changes)

al1img (approved):

PR #252, opened 2025-01-31
Start instance monitoring before instance is launched

mlohvynenko:
ResourceMonitor::ProcessMonitoring() runs on a separate thread and operates with a delay defined by the "poll period" configuration variable.

As a result, ProcessMonitoring() may execute at a moment when instance monitoring has started, but the instance itself has not yet been launched, leading to the following error: `main aos_servicemanager[1659]: (monitoring) Failed to get instance monitoring data: not found (resourceusageprovider.cpp:324)`

### DIFF
--- a/src/common/monitoring/resourcemonitor.cpp
+++ b/src/common/monitoring/resourcemonitor.cpp
@@ -149,7 +149,7 @@ Error ResourceMonitor::StartInstanceMonitoring(const String& instanceID, const I
 
     if (err = mResourceUsageProvider->GetInstanceMonitoringData(
             instanceID, mInstanceMonitoringData.Find(instanceID)->mSecond);
-        !err.IsNone()) {
+        !err.IsNone() && !err.Is(ErrorEnum::eNotFound)) {
         LOG_WRN() << "Can't get instance monitoring data: instanceID=" << instanceID << ", err=" << err;
     }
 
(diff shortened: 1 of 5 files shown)


src/sm/launcher/instance.cpp
```diff
@@ -119,18 +119,24 @@ Error Instance::Start()
         return err;
     }
 
+    if (auto err = SetupMonitoring(); !err.IsNone()) {
+        return AOS_ERROR_WRAP(err);
+    }
+
     auto runStatus = mRunner.StartInstance(mInstanceID, mRuntimeDir, {});
 
+    auto onExit = DeferRelease(&runStatus, [this](runner::RunStatus* runStatus) {
```
al1img, line 128: This is not required because this instance is treated as running and Stop API will be called.

src/common/monitoring/resourcemonitor.cpp
```diff
@@ -476,6 +476,8 @@ void ResourceMonitor::ProcessMonitoring()
             if (auto err = mResourceUsageProvider->GetInstanceMonitoringData(instanceID, instanceMonitoringData);
                 !err.IsNone()) {
                 LOG_ERR() << "Failed to get instance monitoring data: " << err;
+
```
al1img, line 479: But we will have error sometimes?

mykola-kobets-epam (approved):

PR #261, opened 2025-02-13
Improve sm service manager memory consumption

mlohvynenko:


include/aos/common/spaceallocator/spaceallocator.hpp
```diff
@@ -179,7 +179,7 @@ class SpaceAllocator : public SpaceAllocatorItf {
      */
     RetWithError<UniquePtr<SpaceItf>> AllocateSpace(uint64_t size) override
     {
-        return UniquePtr<SpaceItf>(new (&mAllocator) Space(size));
```
al1img, line 182: It should be valid case as well. We have to fix UniquePtr.

tests/common/src/spaceallocator/spaceallocator_test.cpp
```diff
@@ -0,0 +1,42 @@
+/*
```
al1img, line 1: Adding spaceallocator test should be in separate commit

include/aos/common/spaceallocator/spaceallocator.hpp
mlohvynenko: Updated UniquePtr constructors.
But I kept change to use MakeUnique.

al1img (approved):

mykola-kobets-epam (approved):

MykolaSuperman (approved):

mykola-kobets-epam (approved):

PR #272, opened 2025-02-20
[sm,resourcemanager] Reset node config object on load

mlohvynenko:
Make sure that the node config object is reset on load, so all the member arrays are cleared.

### DIFF
--- a/src/sm/resourcemanager/resourcemanager.cpp
+++ b/src/sm/resourcemanager/resourcemanager.cpp
@@ -36,7 +36,7 @@ Error ResourceManager::Init(JSONProviderItf& jsonProvider, HostDeviceManagerItf&
         LOG_ERR() << "Failed to load unit config: err=" << err;
     }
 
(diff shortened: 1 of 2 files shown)


src/sm/resourcemanager/resourcemanager.cpp
```diff
@@ -36,7 +36,7 @@ Error ResourceManager::Init(JSONProviderItf& jsonProvider, HostDeviceManagerItf&
         LOG_ERR() << "Failed to load unit config: err=" << err;
     }
 
-    LOG_DBG() << "Node config version: version=" << mConfig.mVersion << ", error=" << mConfigError;
+    LOG_DBG() << "Node config version: version=" << mConfig->mVersion << ", error=" << mConfigError;
```
al1img, line 39: err=

al1img (approved):

al1img (approved):

MykolaSuperman (approved):

PR #275, opened 2025-02-28
Timer implementation

mykola-kobets-epam:


mlohvynenko (approved):

include/aos/common/tools/algorithm.hpp
```diff
@@ -236,6 +244,41 @@ class AlgorithmItf {
...
+    /**
+     * Finds minimal element using provided comparator.
+     *
+     * @param cmp less comparator.
+     * @return CI.
+     */
+    template <typename Cmp = LessThanComparator<T>>
+    CI Min(Cmp cmp = Cmp()) const
+    {
+        return const_cast<AlgorithmItf&>(*this).Min(cmp);
```
MykolaSolyanko, line 256: could be undefined behaviour if this is a real constant according C++ standard, so this could be an issue in some linters or static analysis

src/common/tools/timer.cpp
```diff
@@ -0,0 +1,158 @@
...
+        for (auto& timer : mRegisteredTimers) {
+            if (timer->mWakeupTime.IsZero()) {
+                break;
+            }
+
+            if (timer->mWakeupTime > cbInvokeThreshold) {
+                break;
+            }
+
+            InvokeTimerCallback(timer);
```
MykolaSolyanko, line 141: So in case of if InvokeTimerCallback takes a lot of time to execute, then another timer will be wait?

src/common/tools/timer.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+            if (timer->mOneShot) {
+                timer->mWakeupTime = Time();
```
MykolaSolyanko, line 144: Why do we need to store a one-shot timer? I mean, I expect to call Create for a one-shot timer and not call Stop, since it is one-shot. So, in this case, when mRegisteredTimers is full and I try to add a new timer, mRegisteredTimers might already contain some one-shot timers. Am I right?

include/aos/common/tools/algorithm.hpp
mykola-kobets-epam: undestood, thank you. I made it vice versa: call const from non-const method & made const_cast conversion of the result

src/common/tools/timer.cpp
MykolaSolyanko: In my opinion, it depends on the types of orchestrators. But that's not the point. I simply expected two approaches for a one-shot timer from a user's perspective. The first approach is that I call the timer and wait for it to execute—and once it has executed, I don’t need to remember to delete it because, being one-shot, the API can manage it automatically. The second case is when I call Stop on the one-shot timer before it triggers. Since many such timers might accumulate, this could lead to issues. One possible solution is either to automatically delete the timer once its callback has finished (i.e., remove it from the pool of registered timers) or to rotate them—in other words, when a new timer is created and there’s no available space, discard those one-shot timers that haven’t executed. I propose that we discuss this.

include/aos/common/tools/timer.hpp
```diff
@@ -123,44 +76,41 @@ class Timer {
...
-            auto err = Stop();
-            if (!err.IsNone()) {
-                return err;
-            }
-
-            err = Create(mInterval, functor, mOneShot, arg);
-            if (!err.IsNone()) {
-                return err;
-            }
+        if (auto err = Create(mInterval, functor, mOneShot, arg); !err.IsNone()) {
```
al1img, line 79: Unclear why we need `Reset` with callback? Should we call stop before create in reset?

include/aos/common/tools/thread.hpp
```diff
@@ -503,14 +503,39 @@ class ThreadPool : private NonCopyable {
...
+        return ErrorEnum::eNone;
+    }
+
+    /**
+     * Adds task to task queue.
+     *
+     * @param functor task functor.
+     * @return Error.
+     */
+    Error AddTask(const StaticFunction<cMaxTaskSize>& functor)
```
al1img, line 527: Why this is required?

include/aos/common/tools/timer.hpp
```diff
@@ -40,80 +35,38 @@ class Timer {
...
      *
      * @param interval timer interval.
+     * @param callback callback.
+     * @param oneShot specifies whether timer should be called exactly once.
+     * @param arg callback argument.
      * @return Error code.
      */
     template <typename F>
-    Error Create(Duration interval, F functor, bool oneShot = true, void* arg = nullptr)
+    Error Create(Duration interval, F callback, bool oneShot = true, void* arg = nullptr)
```
al1img, line 44: Rename to `Start`.

src/common/tools/timer.cpp
```diff
@@ -0,0 +1,158 @@
...
+    auto cmpWakeupTime = [](const Timer* left, const Timer* right) {
+        if (left->mWakeupTime.IsZero()) {
+            return false;
+        }
+
+        if (right->mWakeupTime.IsZero()) {
+            return true;
+        }
+
+        return left->mWakeupTime.UnixNano() < right->mWakeupTime.UnixNano();
```
al1img, line 117: ` return left->mWakeupTime < right->mWakeupTime;` ?

include/aos/common/tools/algorithm.hpp
```diff
@@ -288,7 +331,7 @@ class AlgorithmItf {
      * @param sortFunc sort function.
      * @param tmpValue tmp value used for temporary storage.
      */
-    template <typename F>
+    template <typename F = LessThanComparator<T>>
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/tools/algorithm.hpp
```diff
@@ -309,31 +352,20 @@ class AlgorithmItf {
      * @tparam F type of sort function.
      * @param sortFunc sort function.
      */
-    template <typename F>
-    void Sort(F sortFunc)
+    template <typename F = LessThanComparator<T>>
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/tools/string.hpp
```diff
@@ -319,7 +319,7 @@ class String : public Array<char> {
                 return err;
             }
 
-            for (size_t j = 0; j < Min(oldSubstr.Size(), newSubstr.Size()); j++) {
+            for (size_t j = 0; j < aos::Min(oldSubstr.Size(), newSubstr.Size()); j++) {
```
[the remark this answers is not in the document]
mykola-kobets-epam: otherwise it uses AlgorithmItf::Min

src/common/tools/timer.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        auto min = mRegisteredTimers.Min(cmpWakeupTime);
+        if (min != mRegisteredTimers.end() && !(*min)->mWakeupTime.IsZero()) {
+            mCommonCondVar.Wait(lock, (*min)->mWakeupTime);
+        } else {
+            mCommonCondVar.Wait(lock);
+        }
+    }
+}
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/tools/timer.hpp
```diff
@@ -123,44 +76,41 @@ class Timer {
...
         }
 
         return ErrorEnum::eNone;
     }
 
 private:
-    static constexpr int cTimerSigevNotify = AOS_CONFIG_TIMER_SIGEV_NOTIFY;
+    Duration                                mInterval {};
+    bool                                    mOneShot {};
+    StaticFunction<cDefaultFunctionMaxSize> mFunction;
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/tools/timer.hpp
```diff
@@ -123,44 +76,41 @@ class Timer {
...
-    StaticFunction<cDefaultFunctionMaxSize> mFunction;
+    static void ProcessTimers(void* arg);
+    static void UpdateWakeupTime(const Time& now, Timer* timer);
+    static void InvokeTimerCallback(Timer* timer);
+
+    static StaticArray<Timer*, cMaxTimersCount> mRegisteredTimers;
+    static Mutex                                mCommonMutex;
+    static ConditionalVariable                  mCommonCondVar;
+
+    static Thread<cDefaultFunctionMaxSize, cDefaultThreadStackSize> mManagementThread;
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/config.hpp
```diff
@@ -1026,4 +1026,11 @@
 #define AOS_CONFIG_CLOUDPROTOCOL_BEARER_TOKEN_LEN 256
 #endif
 
+/**
+ * Maximum supported number of timers.
+ */
+#ifndef AOS_CONFIG_TIMERS_MAX_COUNT
+#define AOS_CONFIG_TIMERS_MAX_COUNT 20
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

include/aos/common/tools/thread.hpp
mykola-kobets-epam: It's used by the timer to invoke input callback

include/aos/common/tools/timer.hpp
mykola-kobets-epam: Reset no longer takes arguments

include/aos/common/tools/timer.hpp
mykola-kobets-epam: done

src/common/tools/timer.cpp
mykola-kobets-epam: done

src/common/tools/timer.cpp
mykola-kobets-epam: decided to stop timer in registered callback

al1img (approved):

include/aos/common/tools/timer.hpp
```diff
@@ -53,7 +53,16 @@ class Timer {
 
         LockGuard lock {mMutex};
 
-        if (auto err = mFunction.Capture(callback, arg); !err.IsNone()) {
+        const auto wrappedCallback = [this, callback](void* arg) {
+            callback(arg);
+
+            LockGuard lock {mMutex};
```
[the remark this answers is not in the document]
mykola-kobets-epam: done

PR #276, opened 2025-02-28
Fix variant copy

mlohvynenko:


include/aos/common/tools/variant.hpp
```diff
@@ -263,6 +263,10 @@ class Variant {
      */
     friend Log& operator<<(Log& log, const Variant& variant)
     {
+        if (variant.mTypeIndex == cInvalidTypeIndex) {
```
mykola-kobets-epam, line 266: I think we should still print smth, like "<>" to indicate an empty variant

mykola-kobets-epam (approved):

include/aos/common/tools/variant.hpp
mlohvynenko: nice point

mykola-kobets-epam (approved):

al1img (approved):

PR #283, opened 2025-03-07
[common,tools] Improve error message concatenation logic

MykolaSuperman:
Enhance error message handling to better combine error messages while preventing buffer overflows. The changes ensure that when multiple error messages are present, they are concatenated with a space separator and within the message buffer's size

### DIFF
--- a/include/aos/common/tools/error.hpp
+++ b/include/aos/common/tools/error.hpp
@@ -99,11 +99,27 @@ public:
             mLineNumber = err.mLineNumber;
         }
 
+        if (msg && err.mMessage[0] != '\0') {
+            CopyMessage(msg);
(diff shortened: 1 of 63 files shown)


include/aos/common/tools/error.hpp
```diff
@@ -99,11 +99,27 @@ class Error {
             mLineNumber = err.mLineNumber;
         }
 
+        if (msg && err.mMessage[0] != '\0') {
```
al1img, line 102: We should not concatenate error.  The logic should be as following if `msg` is not `nullptr` and `mMessage` is empty than set `msg` to `mMessage`. Do nothing, otherwise.

PR #290, opened 2025-03-20
Fixes for aos core zephyr application

mlohvynenko:


mykola-kobets-epam (approved):

include/aos/common/tools/fs.hpp
```diff
@@ -595,6 +595,49 @@ class FS {
...
     }
 
+    /**
+     * Calculates size of the file or directory.
+     *
+     * @param path file or directory path.
+     * @param dirIterators storage to store subdirectories iterators.
+     * @return RetWithError<size_t>.
+     */
+    static RetWithError<size_t> CalculateSize(const String& path, Array<DirIterator>& dirIterators)
```
MykolaSuperman, line 621: In my opinion, the interface is a bit unclear; that is, we pass an array (which is mutable) to the invoked function, and then we don't use it. I understand that this is a memory issue, but I think it would be better to do it a bit differently, for example, by passing an allocator as a parameter. It might be even better to add the allocator directly to the class where CalculateSize is used, since it's essentially its internal state.

But if everyone agrees with this implementation, then that's fine, although in my view the interface isn't very intuitive.

include/aos/common/tools/fs.hpp
```diff
@@ -54,6 +55,38 @@ class FS {
...
+         * Move constructor.
+         *
+         * @param other iterator to move from.
+         */
+        DirIterator(DirIterator&& other)
+            : mDir(other.mDir)
+            , mEntry(other.mEntry)
+            , mRoot(other.mRoot)
+        {
+            other.mDir = nullptr;
```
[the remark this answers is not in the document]
mlohvynenko: On my opinion it redundant: we are leaving moved object in a valid state.

include/aos/common/tools/fs.hpp
mlohvynenko: Option #1: pass allocator:
Caller code should allocate enough space, so with this approach I see no benefits, cause you need to know implementation details.

Option #2: add allocator to FS class, make this function non static.
Simplifies interface, only 1 param is  needed - path.
cpp file for FS to be added, FS  object should be properly initialized (on app init level) and passed to caller code as a member, or some kind of singleton
Otherwise, we will create object in a caller code what will likely end up with stack overflow issues.

Option #3: accept mutable dir array ref (current implementation):
Caller code allocates memory, but it make interface less obvious.

MykolaSuperman (approved):

PR #292, opened 2025-03-24
[monitoring] Send alert status=continue when curValue > minTreshold

mykola-kobets-epam:


src/common/monitoring/alertprocessor.cpp
```diff
@@ -160,49 +152,50 @@ Error AlertProcessor::HandleMaxThreshold(uint64_t currentValue, const Time& curr
 
 Error AlertProcessor::HandleMinThreshold(uint64_t currentValue, const Time& currentTime)
 {
-    Error err = ErrorEnum::eNone;
+    if (currentValue > mMinThreshold) {
```
al1img, line 155: Ideally should be '>='.

MykolaSolyanko (approved):

src/common/monitoring/alertprocessor.cpp
mykola-kobets-epam: fixed

al1img (approved):

PR #305, opened 2025-04-29
[resourcemonitor] Move subscribe to start

MykolaSuperman:


mlohvynenko (approved):

src/common/monitoring/resourcemonitor.cpp
```diff
@@ -59,6 +55,10 @@ Error ResourceMonitor::Start()
 {
     LOG_DBG() << "Start monitoring";
 
+    if (auto err = mConnectionPublisher->Subscribe(*this); !err.IsNone()) {
```
al1img, line 58: What issue do you see with making subscription in the init?

MykolaSuperman: If we call Stop, and don't move Subscribe to start, then after Stop and when we call Start again, we cannot subscribe to events. And there could be a possible issue where the event could be sent (if we left it in Init but we still haven't started).

al1img (approved):

PR #316, opened 2025-05-21
[all] Use crypto random interface to generate UUID

mlohvynenko:


src/iam/permhandler/permhandler.cpp
```diff
@@ -108,16 +121,22 @@ InstancePermissions* PermHandler::FindByInstanceIdent(const InstanceIdent& insta
...
+    auto [uuid, err] = uuid::CreateUUID(*mRandomProvider);
+    if (!err.IsNone()) {
+        return {{}, AOS_ERROR_WRAP(err)};
+    }
 
-    } while (FindBySecret(newSecret) != mInstancesPerms.end());
+    result = uuid::UUIDToString(uuid);
 
-    return newSecret;
+    return result;
```
MykolaSuperman, line 139: maybe just 
```
return uuid::UUIDToString(uuid);
```

MykolaSuperman (approved):

src/iam/permhandler/permhandler.cpp
mlohvynenko: uuid static string has less buffer size then secret, so compilation fails with the proposed approach
but works fine with an explicit constructor call

`return StaticString<cSecretLen>(uuid::UUIDToString(uuid));`

include/aos/common/tools/uuid.hpp
```diff
@@ -11,6 +11,10 @@
 #include "aos/common/config.hpp"
 #include "aos/common/tools/string.hpp"
 
+namespace aos::crypto {
+class RandomItf;
```
al1img, line 15: Why not just include "crypto.hpp"?

include/aos/common/tools/uuid.hpp
```diff
@@ -31,9 +35,9 @@ using UUID = StaticArray<uint8_t, cUUIDSize>;
 /**
  * Creates unique UUID.
  *
- * @return UUID.
+ * @return RetWithError<UUID>.
  */
-UUID CreateUUID();
+RetWithError<UUID> CreateUUID(class crypto::RandomItf& randomProvider);
```
al1img, line 40: We have `CreateUUIDv5` crypto provider method. 

I suggest moving this method there as well. In any case user CreateUUID requires additional interface (randomitf or cryptoprovideritf). Each crypto provider has its own mechanism to generate UUID. Let's use this approach.

mlohvynenko: Make sense, but enhancing this interface doesn't seem right to me, at least from ProviderItf doc.
Probably adding UUIDProviderItf and implementing it by OpenSSLCryptoProvider/MbedTLSCryptoProvider classes 
is a better approach

```
/**
 * Provides interface to manage certificate requests.
 */
class ProviderItf {
public:
```

include/aos/common/tools/uuid.hpp
mykola-kobets-epam: probably better create a dedicated hpp for RandomItf

mykola-kobets-epam (approved):

MykolaSuperman (approved):

al1img (approved):

PR #336, opened 2025-07-01
Add download item

MykolaSuperman:
In some downloads, additional data is required, such as the target type, target ID, and version, which should be passed through alerts.


mlohvynenko (approved):

include/aos/common/downloader/downloader.hpp
```diff
@@ -8,30 +8,45 @@
...
+    /**
+     * Constructor.
+     *
+     * @param url URL.
+     * @param path path to file.
+     * @param targetType target type.
+     * @param targetID target ID.
+     * @param version version.
+     */
+    DownloadItem(const StaticString<cURLLen>& url, const StaticString<cFilePathLen>& path,
```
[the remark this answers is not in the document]
al1img: use const String for now.

include/aos/common/downloader/downloader.hpp
```diff
@@ -8,30 +8,45 @@
...
 /**
- * Download content type.
+ * Download item.
  */
-class DownloadContentType {
-public:
-    enum class Enum { eService, eLayer };
-
-    static const Array<const char* const> GetStrings()
+struct DownloadItem {
```
al1img, line 21: Not sure we need additional struct for 5 params. We don't have clear rule about max function params count. But for me more readable to call this particular method with 5 params then fill DownloadItem struct before.

include/aos/common/downloader/downloader.hpp
```diff
@@ -40,12 +23,12 @@ class DownloaderItf {
     /**
      * Downloads file.
      *
-     * @param url URL.
-     * @param path path to file.
-     * @param contentType content type.
+     * @param item download item.
```
mlohvynenko, line 26: Please update doc comment
it looks outdated

mykola-kobets-epam (approved):

PR #366, opened 2025-09-08
sm: networkmanager: fix remove instance

MykolaSuperman:
Previously, when deleting a network we iterated over all its instances and called a cleanup function for each. That function also cleared the cache, which removed the instance object from the map we were iterating. This caused undefined behavior in the loop. The cleanup logic has been adjusted to avoid modifying the container while iterating.


src/sm/networkmanager/networkmanager.cpp
```diff
@@ -261,6 +261,32 @@ Error NetworkManager::RemoveInstanceFromNetwork(const String& instanceID, const
 
     Error err;
 
+    if (auto errClenupResources = CleanupInstanceNetworkResources(instanceID, networkID);
```
mykola-kobets-epam, line 264: errClenup* => errCleanup

mykola-kobets-epam (approved):

al1img (approved):

src/sm/networkmanager/networkmanager.cpp
MykolaSuperman: Done

PR #371, opened 2025-09-17
Update types

al1img:


mykola-kobets-epam (commented):
mb rename it to EnvVars, "array" suffix looks similar to hungarian notation

PR #377, opened 2025-09-26
common: alerts: implement module

mlohvynenko:


mykola-kobets-epam (approved):

src/core/cm/alerts/alerts.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+        return *alertCopy == item;
+    }) != mAlerts.end();
+}
+
+Error Alerts::CacheAlert(const cloudprotocol::AlertVariant& alert)
+{
+    if (mAlerts.IsFull()) {
+        return AOS_ERROR_WRAP(ErrorEnum::eNoMemory);
+    }
```
MykolaSuperman, line 193: but this case handle `EmplaceBack`

src/core/cm/alerts/alerts.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    package->mItems.Assign(Array<cloudprotocol::AlertVariant>(mAlerts.begin(), count));
+
+    return package;
+}
+
+void Alerts::ShrinkCache(size_t count)
+{
+    for (size_t i = 0; i < count && !mAlerts.IsEmpty(); i++) {
+        mAlerts.Erase(mAlerts.begin());
```
MykolaSuperman, line 212: maybe we could use this API?
`virtual I Erase(CI first, CI last) = 0;`

MykolaSuperman (approved):

src/core/cm/alerts/alerts.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Alert packages count.
+ */
+constexpr auto cAlertPackagesCount = AOS_CONFIG_CM_ALERTS_ALERT_PACKAGES_COUNT;
+
+/**
+ * Alerts.
+ */
+class Alerts : public ReceiverItf, public ConnectionListenerItf {
```
al1img: private ConnectionListenerItf

src/core/cm/alerts/alerts.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (IsDuplicated(alert)) {
+        ++mDuplicatedAlerts;
+
+        return ErrorEnum::eNone;
+    }
+
+    if (auto err = mAlerts.EmplaceBack(alert); !err.IsNone()) {
+        ++mSkippedAlerts;
+
+        return err;
```
al1img: Need to handle this case more gentle, otherwise we will display error in the log for each new alert in offline mode. So, it should be done in the following way:

1. if err is not "no memory" than we should return err.
2. Otherwise, we should display warning (not error) only once, then if no error, reset this state.

src/core/cm/alerts/alerts.md
```diff
@@ -27,3 +27,29 @@ classDiagram
 
     Alerts ..> SenderItf
 ```
+
```
al1img: Need to describe how alert messages are aggregated and handled in offline mode.

src/core/cm/alerts/alerts.cpp
mlohvynenko: Only no memory is possible here
The initial plan was to handle the error on the caller side.

Though, if we already increase mSkippedAlerts, there's no need to return error at all in case of buffer overflow

MykolaSuperman (approved):
Reviewed-by: Mykola Solianko <mykola_solianko@epam.com>

mykola-kobets-epam (approved):

PR #382, opened 2025-10-07
Crypto refactoring

mykola-kobets-epam:


src/core/common/crypto/itf/crypto.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include "hash.hpp"
+#include "privkey.hpp"
+#include "x509.hpp"
+
+namespace aos::crypto {
+
+/**
+ * Random generator interface.
+ */
+class RandomItf {
```
mlohvynenko, line 22: shouldn't RandomItf be moved to a dedicated header (like HashItf for example)

mlohvynenko (approved):

PR #434, opened 2025-12-06
cm: launcher: fix compilation issues

mlohvynenko:
This patch fixes the next compilation issues:
- aplies blob info provider interface changes;
- adds update item type to instance ident constructor calls.


MykolaSuperman (approved):

PR #516, opened 2026-02-23
cm: monitoring: omit duplicate states in instance status

mlohvynenko:
This patch ensures that duplicate states are not added if the new state is the same as the last one.

### DIFF
--- a/src/core/cm/monitoring/monitoring.cpp
+++ b/src/core/cm/monitoring/monitoring.cpp
@@ -177,6 +177,10 @@ void Monitoring::OnInstancesStatusesChanged(const Array<InstanceStatus>& statuse
             it->mNodeID                      = status.mNodeID;
         }
 
+        if (!it->mStates.IsEmpty() && it->mStates.Back().mState == status.mState) {
+            continue;
+        }
+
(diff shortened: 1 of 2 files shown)


src/core/cm/monitoring/monitoring.cpp
```diff
@@ -177,6 +177,10 @@ void Monitoring::OnInstancesStatusesChanged(const Array<InstanceStatus>& statuse
             it->mNodeID                      = status.mNodeID;
         }
 
+        if (!it->mStates.IsEmpty() && it->mStates.Back().mState == status.mState) {
```
al1img, line 180: I think we need to fix root cause as well.

mykola-kobets-epam (approved):

MykolaSuperman (approved):

src/core/cm/monitoring/monitoring.cpp
mlohvynenko: Discussed on daily, the root cause might be SM reboot
Anyway, this fix fulfills a missed requirement to skip duplicates.

al1img (approved):

PR #525, opened 2026-03-04
sm: networkmanager: fix iterator invalidation in RemoveNetworks

MykolaSuperman:
Fixed critical use-after-erase bug where iterator was accessed after Erase() call. Also fixed race condition by moving element removal before unlock to prevent iterator invalidation by other

### DIFF
--- a/src/core/sm/networkmanager/networkmanager.cpp
+++ b/src/core/sm/networkmanager/networkmanager.cpp
@@ -997,27 +997,31 @@ Error NetworkManager::RemoveNetworks(const Array<aos::NetworkParameters>& networ
     for (auto it = mNetworkProviders.begin(); it != mNetworkProviders.end();) {
         auto itNetwork = networks.FindIf([&](const auto& network) { return it->mFirst == network.mNetworkID; });
         if (itNetwork == networks.end()) {
+            // Save data before erase to avoid use-after-erase and minimize lock time
+            auto networkID   = it->mFirst;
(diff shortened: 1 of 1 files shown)


mlohvynenko (approved):

src/core/sm/networkmanager/networkmanager.cpp
```diff
@@ -997,27 +997,31 @@ Error NetworkManager::RemoveNetworks(const Array<aos::NetworkParameters>& networ
     for (auto it = mNetworkProviders.begin(); it != mNetworkProviders.end();) {
         auto itNetwork = networks.FindIf([&](const auto& network) { return it->mFirst == network.mNetworkID; });
         if (itNetwork == networks.end()) {
+            // Save data before erase to avoid use-after-erase and minimize lock time
+            auto networkID   = it->mFirst;
+            auto networkInfo = it->mSecond;
+
+            it = mNetworkProviders.Erase(it);
```
mykola-kobets-epam, line 1004: May be better erase in the end of the loop clause?

mykola-kobets-epam (approved):

src/core/sm/networkmanager/networkmanager.cpp
MykolaSuperman: it will be deadlock because it call RemoveNetwork which also use lock, but it will posiible refactoring in the next version according new design

al1img (approved):

PR #527, opened 2026-03-05
build: add cmake configure step to linting

mlohvynenko:
compile_commands.json is generated during cmake configure step, so it should be run before cppcheck.
This patch makes it possible to execute lint step without building the project, which fixes the issue with linting on CI.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -55,7 +55,9 @@ conan_setup() {
     conan install ./conan/ --output-folder build --settings=build_type="$ARG_BUILD_TYPE" --build=missing
 }
 
-build_project() {
+cmake_configure() {
+    conan_setup
+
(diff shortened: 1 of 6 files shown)


mykola-kobets-epam (approved):

build.sh
```diff
@@ -118,7 +124,6 @@ build_target() {
         clean_build
     fi
 
-    conan_setup
```
al1img, line 121: Why is it removed?

mlohvynenko: Not needed, as this action is required for the configure step, and cmake_configure calls conan_setup on the first line

al1img (approved):

MykolaSuperman (approved):

PR #529, opened 2026-03-09
sm: launcher: handle instances offline TTL

mlohvynenko:


src/core/sm/launcher/launcher.cpp
```diff
@@ -345,6 +382,127 @@ void Launcher::RunRebootThread()
...
+            });
+            !err.IsNone()) {
+            LOG_ERR() << "Failed to schedule stop for expired TTL instance" << Log::Field("instance", instance.mInfo)
+                      << Log::Field(AOS_ERROR_WRAP(err));
+        }
+    }
+
+    lock.Unlock();
+
+    if (auto err = mOfflineTTLPool.Wait(); !err.IsNone()) {
```
MykolaSuperman, line 485: Is it possible to have a race here? I mean, on line `458` we capture instance and then unlock the mutex, but `UpdateInstances` could modify `mInstances` in the meantime.

al1img (approved):

src/core/sm/launcher/launcher.cpp
mlohvynenko: Fixed potential race by integrating mLaunchInProgress flag logic with the offline TTL handler timer invocation

mykola-kobets-epam (approved):

src/core/sm/launcher/launcher.cpp
MykolaSuperman: Is loks like there is still a race in the offline TTL path: StopExpiredInstances() captures &instance from mInstances in the task lambda (launcher.cpp:470), then releases mMutex before waiting for the pool (launcher.cpp:483).
While the lock is released, OnInstancesStatusesReceived() (launcher.cpp:159) can call RemoveInstanceData() (launcher.cpp:525), which erases elements from mInstances via RemoveIf() (launcher.cpp:1000)

mlohvynenko: Agree with this rare case
Actually, this brach covers a case when the runtime notifies of some outdated component on module start.
Rootfs/boot -> on successful installation they reboot and sent inactive state for the prev version (so Launcher can clear data from DB, to prevent further start attempts for the removed item).

Fixes this case by not mutating mInstances, but simply deleting from DB (anyway, the component will not be in mInstances for this scenario)

MykolaSuperman (approved):

PR #534, opened 2026-03-23
Checksum imagemanager

MykolaSuperman:


src/core/common/tools/fs.hpp
```diff
@@ -260,6 +260,7 @@ struct FileInfo {
     FileInfo() = default;
 
     StaticArray<uint8_t, crypto::cSHA256Size> mSHA256;
```
mlohvynenko: Probably we should store AlgEnum value and a static array with the ectual hash (using Max(crypto::cSHA256Size, ....) as size if needed) >

mlohvynenko (approved):

src/core/common/tools/fs.cpp
```diff
@@ -113,7 +113,12 @@ Error FileInfoProvider::GetFileInfo(const String& path, FileInfo& info)
 
     info.mSize = size;
 
-    return crypto::CalculateFileHash(path, crypto::HashEnum::eSHA256, *mHashProvider, info.mSHA256);
+    if (auto hashErr = crypto::CalculateFileHash(path, crypto::HashEnum::eSHA256, *mHashProvider, info.mSHA256);
```
al1img: It is not efficient to calculate checksum twice. Add bool parameter sha3 with default set to false.

mykola-kobets-epam (approved):

al1img (approved):

PR #539, opened 2026-03-26
cm: launcher: force balancing if any not scheduled instance on start detected

mykola-kobets-epam:


mlohvynenko (approved):

src/core/cm/launcher/instance.cpp
```diff
@@ -234,6 +238,32 @@ Error Instance::SetActive(const String& nodeID, const String& runtimeID)
...
 }
 
+Error Instance::SetDefaultRuntimes()
+{
+    static const char* cDefaultRuntimes[] = {
+        "crun",
+        "runc",
+    };
+
+    assert(mItemConfig);
```
al1img: Do not use assert, we will get rid of it soon. Better to return error.

al1img (approved):

src/core/cm/launcher/instance.cpp
mykola-kobets-epam: fixed

PR #541, opened 2026-04-02
sm: launcher: rename SendUpdateInstancesStatuses

mlohvynenko:
This patch renames the SendUpdateInstancesStatuses method to SendInstanceStatus with one instance status.


mykola-kobets-epam (approved):

PR #571, opened 2026-05-21
cm: launcher: rank scheduling errors in balancer

mykola-kobets-epam:


src/core/cm/launcher/balancer.cpp
```diff
@@ -100,63 +100,66 @@ Error Balancer::PerformNodeBalancing(Array<SharedPtr<Instance>>& instances)
             continue;
         }
 
-        Error scheduleErr = ErrorEnum::eNotFound;
+        RankedError scheduleErr;
```
al1img, line 103: Still do not understand why we need ranked error. Instance scheduling should fail on first error.
In our case, if there are different manifests for different arch. First we should check if there are nodes matches item arches. If not, return error - no suitable node architecture found. Then we try to schedule according to resource, if there are not enough resources available, return resource error.

PR #572, opened 2026-05-22
cm: launcher: "not enough resources" error should have higher priority

mykola-kobets-epam:
There are two main error types in balancer:
- not enough resources: CPU, RAM, max instance num
- no matching node/runtime


src/core/cm/launcher/balancer.hpp
```diff
@@ -56,6 +56,12 @@ class Balancer {
 
 private:
     using NodeRuntimes = StaticMap<Node*, StaticArray<const RuntimeInfo*, cMaxNumNodeRuntimes>, cMaxNumInstances>;
+    enum class ScheduleDecision { eScheduled, eSkipManifest, eError };
```
al1img, line 59: Use Aos enum to properly display it in log.
What does eSkipManifest mean?

PR #581, opened 2026-06-02
sm: launcher: integrate network manager

mlohvynenko:


src/core/sm/launcher/itf/instanceidprovider.hpp
```diff
@@ -0,0 +1,39 @@
...
+    virtual ~InstanceIDProviderItf() = default;
+
+    /**
+     * Returns instance ID.
+     *
+     * @param instance instance ident.
+     * @param[out] instanceID instance ID.
+     * @return Error.
+     */
+    virtual Error GetInstanceID(const InstanceIdent& instance, String& instanceID) const = 0;
```
mykola-kobets-epam, line 34: Intention is not really clear. as instanceID is a part of ident. As the result id is going to be used with network interface, probably better rename it to networkID

mykola-kobets-epam (approved):

mykola-kobets-epam (approved):

src/core/sm/launcher/itf/instanceidprovider.hpp
mlohvynenko: it's discussible, launcher runtime container uses instance id as well - to create container runtime folders at least 

The best approach is to pass InstanceIdent into NetworkManagerItf instead of this instanceID.
But it will be done in future.

MykolaSuperman (approved):

al1img (approved):

PR #590, opened 2026-06-16
cm: launcher: fix instance status

mykola-kobets-epam:


mlohvynenko (approved):

src/core/cm/launcher/instancemanager.cpp
```diff
@@ -187,18 +187,25 @@ Error InstanceManager::UpdateStatus(const InstanceStatus& status)
...
+            firstErr = err;
         }
+    } else {
+        LOG_WRN() << "Received status for unknown instance"
+                  << Log::Field("instance", static_cast<const InstanceIdent&>(status));
+    }
 
-        // Not expected instance received from SM.
-        return AOS_ERROR_WRAP(ErrorEnum::eNotFound);
+    if (status.mState != aos::InstanceStateEnum::eInactive) {
```
MykolaSuperman, line 202: `UpdateStatus()` now updates mRunningInstances only when `status.mState != eInactive`. That means if a running instance was previously added to `mRunningInstances` and later SM sends `OnInstanceStatusReceived(... eInactive ...)`, the active Instance status is updated, but the old running entry remains. `Node::ResendInstances()` filters `mRunningInstances` for non-inactive entries, so future resend/balancing logic can still believe the stopped instance is running and skip a needed restart or produce wrong stop/start decisions. The inactive path should remove the matching running entry, likely by ident/version plus node/runtime.

mykola-kobets-epam: point is valid, added else clause which removes inactive instance status from runing-instances list.

al1img (approved):

PR #593, opened 2026-06-18
Fix imagemanager allocation

MykolaSuperman:


src/core/common/spaceallocator/spaceallocator.hpp
```diff
@@ -157,6 +157,42 @@ class Partition {
...
 
+    /**
+     * Adjusts allocated size without changing allocation count.
+     * Used by resize operations so that mAllocationCount tracks live Space objects only.
+     *
+     * @param oldSize previously allocated size to return.
+     * @param newSize new size to reserve.
+     * @return Error.
+     */
+    Error AdjustSize(size_t oldSize, size_t newSize)
```
al1img, line 168: Rename method to `Resize`.  It looks like we can keep only one parameter as `deltaSize`. It could be positive and negative in the general case. In case negative delta size we can just update `mAvailableSize` for positive delta size we shuould check if there is enough available space and try to remove outdated items.

PR #596, opened 2026-06-24
Imagemanager used blobs

al1img:


src/core/sm/imagemanager/imagemanager.cpp
```diff
@@ -1099,6 +1176,36 @@ Error ImageManager::HandleItemsIntegrity()
     return ErrorEnum::eNone;
 }
 
+Error ImageManager::AddInstallingItems(
+    Array<StaticString<cFilePathLen>>& usedBlobs, Array<StaticString<cFilePathLen>>& usedLayers)
+{
+    for (const auto& installingItem : mInstallingItems) {
```
MykolaSuperman, line 1185: should we use mutex in here when we iterate `mInstallingItems`

src/core/sm/imagemanager/imagemanager.cpp
```diff
@@ -509,15 +471,23 @@ Error ImageManager::DownloadBlob(const String& path, const String& digest, size_
...
-    if (waitInProgress) {
-        if (auto err = WaitForInProgressBlob(descriptor.mDigest); !err.IsNone()) {
+    if (waitInstalling) {
+        if (auto err = WaitForInstallingBlob(descriptor.mDigest); !err.IsNone()) {
             return err;
         }
 
-        auto releaseInProgress
-            = DeferRelease(&descriptor.mDigest, [&](const String* digest) { ReleaseInProgressBlob(*digest); });
+        auto releaseInstalling
```
MykolaSuperman: `releaseInstalling` goes out of scope at the closing `}` of `if (waitInstalling)`, befoe `DownloadBlob` is called — so the blob is removed from `mInstallingBlobs` immediately, and another thread can start downloading the same blob concurrently, so this is expected ?

mykola-kobets-epam (approved):

mlohvynenko (approved):

src/core/sm/imagemanager/imagemanager.cpp
al1img: No, AddInstallingItems is called from RemoveOrphans and RemoveOrphans always called under lock: from RemoveItem and ProcessOutdatedItems.

src/core/sm/imagemanager/imagemanager.cpp
al1img: Good catch. Fixed in additional commit.

MykolaSuperman (approved):

PR #609, opened 2026-07-09
Add wait flag to Timer::Stop

mykola-kobets-epam:


mlohvynenko (approved):

src/core/cm/alerts/alerts.cpp
```diff
@@ -102,7 +102,7 @@ Error Alerts::Stop()
         err = AOS_ERROR_WRAP(unsubscribeErr);
     }
 
-    if (auto stopErr = mSendTimer.Stop(); !stopErr.IsNone()) {
+    if (auto stopErr = mSendTimer.Stop(true); !stopErr.IsNone()) {
```
mlohvynenko: using a dedicated enum would be better, cause it`s unclear for usages what does true mean

src/core/common/tools/thread.hpp
```diff
@@ -662,8 +670,11 @@ class ThreadPool : private NonCopyable {
 
     /**
      * Shutdowns all pool threads.
+
+     * @param waitAll wait for all tasks to be finished.
+     * @return Error.
      */
-    Error Shutdown()
+    Error Shutdown(bool waitAll = true)
```
al1img: There is dedicated method `Wait` that should be called before shutdown If need to wait all task to be finished before shutdown.

src/core/cm/alerts/alerts.cpp
mykola-kobets-epam: done

src/core/common/tools/thread.hpp
mykola-kobets-epam: I split shutdown and wait, and wait now should be called after shutdown

al1img (approved):

MykolaSuperman (approved):

mlohvynenko (approved):

PR #610, opened 2026-07-09
sm: launcher: fix instance data invalidation on remove

al1img:
RemoveInstances scheduled each instance removal as a concurrent task on the launch pool, and every task captured a raw InstanceData* pointing into mInstances. RemoveInstance erased its own entry from mInstances as soon as it finished, which shifts trailing elements down to fill the gap. Other tasks still running concurrently held pointers into those now-shifted (or destructed) slots, so they ended up reading and writing the wrong instance, or freed memory.

Split the work instead: tasks now only release external resources (storage entry, network) via the renamed ReleaseInstance, without touching mInstances. The actual removal from mInstances happens in a single serial pass after mLaunchPool.Wait(), under mMutex, once no task can be holding a pointer into the array anymore. Each entry is re-checked for the eInactive state at removal time, so instances that failed to stop or failed release (now marked eFailed) are left in place instead of being purged.

### DIFF
--- a/src/core/sm/launcher/launcher.cpp
+++ b/src/core/sm/launcher/launcher.cpp
@@ -1112,7 +1112,7 @@ RetWithError<Launcher::InstanceData*> Launcher::AddInstanceData(const InstanceIn
     return itInstance;
 }
 
-Error Launcher::RemoveInstance(InstanceData& instanceData)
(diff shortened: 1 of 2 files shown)


mlohvynenko (approved):

mykola-kobets-epam (approved):

PR #613, opened 2026-07-11
sm: launcher: run instance network start/stop as pool tasks

al1img:
Move network setup/teardown out of StartInstance/StopInstance and into new StartNetworks/StopNetworks helpers that queue StartInstanceNetwork/ StopInstanceNetwork calls on the launch thread pool, so networks are started/stopped in parallel rather than serialized with the instance start/stop itself.

Add StopAllNetworks, invoked from Stop() after StopAllInstances, to tear down networks for all service instances on launcher shutdown, and run it via the launch pool like the other batch operations.

Tag the UpdateInstancesImpl and instance/network start/stop log messages with a "[profiling]" marker to make timing of these stages easier to trace.

### DIFF
--- a/src/core/sm/launcher/launcher.cpp
+++ b/src/core/sm/launcher/launcher.cpp
@@ -115,10 +115,21 @@ Error Launcher::Stop()
 
         lock.Unlock();
 
-        StopAllInstances();
+        auto err = mLaunchPool.Run();
+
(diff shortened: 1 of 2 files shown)


mlohvynenko (approved):

MykolaSuperman (approved):

PR #615, opened 2026-07-13
sm: networkmanager: guard concurrent last-instance network teardown

MykolaSuperman:
RemoveInstances releases instances as parallel launch-pool tasks, so several ReleaseInstanceNetwork calls for the last instances on a shared network can run concurrently. Each removes its own entry from mInstanceNetworkInfos first, so both then observe no instances left and each tears the network down (storage RemoveNetworkInfo + CM ReleaseNodeNetwork), the second failing with "not found".

Claim the teardown atomically via mNetworkProviders under the lock: only the release that still finds the network present removes it and proceeds to clean up storage and the CM.


mlohvynenko (approved):

al1img (approved):

mykola-kobets-epam (approved):

PR #634, opened 2026-07-29
sm: launcher: implement InitInstances

al1img:
At startup a runtime only knows what it is currently running, while the launcher knows the full set of instances that should exist. Each runtime must reconcile its own state against that list: stop anything not in it and properly initialize already-running instances, without starting anything new yet.

Group the stored instances by runtime and call each registered runtime's InitInstances with only the instances that belong to it. Add the corresponding InitInstances mock to RuntimeMock and cover the grouping behavior with a unit test. Grow the launcher allocator to fit the extra per-runtime instances array.


mykola-kobets-epam (approved):

mlohvynenko (approved):

MykolaSuperman (approved):

PR #636, opened 2026-07-31
Redesign allocator

al1img:


src/core/common/tools/memory.hpp
```diff
@@ -389,30 +302,189 @@ class UniquePtr : private NonCopyable {
...
+protected:
+    /**
+     * Allocator the control block itself was allocated from.
+     */
+    AllocatorItf& mAllocator;
+
+private:
+    virtual void Dispose() = 0;
+
+    size_t mRefCount = 1;
```
MykolaSuperman, line 392: [P1] SharedPtr is no longer thread-safe

The previous implementation serialized TakeAllocation and GiveAllocation using the allocator mutex. SharedControlBlock now increments and decrements the plain size_t mRefCount without synchronization. Copying and destroying SharedPtr instances from different threads is a data race and can lose reference-count updates, causing premature disposal, double disposal, or use-after-free. Please make the reference count atomic with appropriate memory ordering, or protect Take and Give with a mutex.

src/core/sm/launcher/launcher.cpp
```diff
@@ -159,8 +167,15 @@ Error Launcher::UpdateInstances(const Array<InstanceIdent>& stopInstances, const
     // Wait in case previous request is not yet finished
     mThread.Join();
 
-    auto stop  = MakeShared<StaticArray<InstanceIdent, cMaxNumInstances>>(&mAllocator, stopInstances);
-    auto start = MakeShared<InstanceInfoArray>(&mAllocator, startInstances);
+    auto stop = MakeShared<StaticArray<InstanceIdent, cMaxNumInstances>>(mAllocator, stopInstances);
+    if (!stop) {
```
MykolaSuperman, line 178: [P1] Finish the launch on allocation failure

StartLaunch sets mLaunchInProgress before these allocations. If either MakeShared call fails, the return bypasses FinishLaunch, so subsequent updates return eWrongState and Stop can wait forever for mLaunchInProgress to clear. Please call FinishLaunch on both OOM paths, use a scope guard, or allocate before marking the launch as in progress.

src/core/common/pkcs11/pkcs11.cpp
```diff
@@ -550,7 +552,10 @@ RetWithError<SharedPtr<SessionContext>> LibraryContext::PKCS11OpenSession(SlotID
         return {nullptr, static_cast<int>(rv)};
     }
 
-    auto session = MakeShared<SessionContext>(&mAllocator, handle, mFunctionList);
+    auto session = MakeShared<SessionContext>(mAllocator, handle, mFunctionList);
+    if (!session) {
```
MykolaSuperman, line 556: [P2] Close the PKCS#11 session when allocation fails

C_OpenSession has already created a session handle at this point. If MakeShared fails, the function returns without calling C_CloseSession, and the SessionContext destructor that normally closes the handle is never created. This leaks a token session on every allocation failure and can eventually exhaust the token session limit. Please close handle before returning eNoMemory.

src/core/common/tests/crypto/providers/mbedtlsfactory.hpp
```diff
@@ -129,6 +130,7 @@ class MBedTLSCryptoFactory : public CryptoFactoryItf {
     Error Encrypt(const crypto::RSAPublicKey& pubKey, const Array<uint8_t>& msg, Array<uint8_t>& cipher) override;
 
 private:
+    HeapAllocator         mAllocator;
```
mlohvynenko: Why don't you pass AllocatorItf in Init for openssl and mbeedtls factories?

mykola-kobets-epam (approved):

src/core/cm/launcher/instancemanager.cpp
```diff
@@ -549,18 +561,21 @@ RetWithError<SharedPtr<Instance>> InstanceManager::CreateInstance(const Instance
     switch (info.mInstanceIdent.mType.GetValue()) {
     case UpdateItemTypeEnum::eService:
         newInstance = MakeShared<ServiceInstance>(
-            &mAllocator, info, mUIDPool, mGIDPool, *mStorage, mStorageState, mImageInfoProvider, mInstanceAllocator);
+            mAllocator, *mAllocator, info, mUIDPool, mGIDPool, *mStorage, mStorageState, mImageInfoProvider);
```
mykola-kobets-epam, line 564: the same allocator provided twice in the param list

src/core/cm/launcher/instancemanager.cpp
```diff
@@ -549,18 +561,21 @@ RetWithError<SharedPtr<Instance>> InstanceManager::CreateInstance(const Instance
...
         break;
 
     case UpdateItemTypeEnum::eComponent:
-        newInstance
-            = MakeShared<ComponentInstance>(&mAllocator, info, *mStorage, mImageInfoProvider, mInstanceAllocator);
+        newInstance = MakeShared<ComponentInstance>(mAllocator, *mAllocator, info, *mStorage, mImageInfoProvider);
```
mykola-kobets-epam, line 568: the same allocator provided twice in the param list

src/core/sm/launcher/launcher.cpp
al1img: Fixed, see: "tools: make SharedPtr control block ref count thread safe" commit. Currently, I use mutex to protect ref counter.

src/core/common/pkcs11/pkcs11.cpp
al1img: Fixed, see: "common: pkcs11: close session on SessionContext alloc failure" commit.

src/core/common/tests/crypto/providers/mbedtlsfactory.hpp
al1img: Fixed.

src/core/cm/launcher/instancemanager.cpp
al1img: It is not the same allocator provided twice: first parameter is is MakeShared first param - allocator to allocate ServiceInstance. Second allocator is passed to ServiceInstance constructor as parameter.

mlohvynenko (approved):

src/core/cm/launcher/instancemanager.cpp
mykola-kobets-epam: ok

MykolaSuperman (approved):

PR #639, opened 2026-08-11
cm: imagemanager: allocate remaining size for partial downloads

MykolaSuperman:
PrepareDownloadSpace freed the space taken by a partially downloaded blob and then allocated the full blob size, reserving the partial size twice and briefly offering space still occupied on disk to other allocators sharing the partition. Reserve only the bytes still to be written and, on cancel, return the reserved but not written part instead of reallocating the difference, which underflowed when the downloader restarted the file.

Keep the reservation in DownloadSpace, settle it in a single place so the download file is removed before its space is freed, and guard EnsureBlob against paths leaving the space neither accepted nor released. Drop AllocateSpaceForPartialDownloads: the allocator already accounts files present on disk.


al1img (approved):

mykola-kobets-epam (approved):

PR #641, opened 2026-08-11
cm: imagemanager: limit number of stored item versions

MykolaSuperman:
CM kept every removed item version in storage until updateItemTTL expired (30d by default), while all read buffers were sized against cMaxNumItemVersions (2). On the third consecutive update of the same item GetItemInfos() overflowed its array and the launcher failed with "can't add item info", leaving the instance unable to start.

Trim the oldest removed versions before adding a new one, the way SM does it in RemoveOldItemVersions(), so an item never exceeds cMaxNumItemVersions rows. An installed version is never force removed: if no removed version is left to reclaim, the trim just stops.

Move cMaxNumItemVersions to the storage interface as it now describes a storage invariant, and add cMaxNumStoredItems for the whole table. GetAllItemsInfos() buffers counted items instead of rows and would have overflowed the same way past 64 rows, so size them accordingly.

### DIFF
--- a/src/core/cm/imagemanager/imagemanager.cpp
+++ b/src/core/cm/imagemanager/imagemanager.cpp
@@ -34,7 +34,7 @@ Error ImageManager::Init(AllocatorItf& allocator, const Config& config, StorageI
     mFileInfoProvider          = &fileInfoProvider;
     mOCISpec                   = &ociSpec;
 
-    auto items = MakeUnique<StaticArray<ItemInfo, cMaxNumUpdateItems>>(mAllocator);
+    auto items = MakeUnique<StaticArray<ItemInfo, cMaxNumStoredItems>>(mAllocator);
     if (!items) {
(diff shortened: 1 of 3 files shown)


src/core/cm/imagemanager/itf/storage.hpp
```diff
@@ -12,6 +12,16 @@
 
 namespace aos::cm::imagemanager {
 
+/**
+ * Max number of versions stored per update item.
+ */
+constexpr auto cMaxNumItemVersions = 2;
```
al1img: Why do you need it in public? It could be private consts.

MykolaSuperman: to calculate this one 
constexpr auto cMaxNumStoredItems = cMaxNumUpdateItems * cMaxNumItemVersions;

in this case than i need to repeat in here

al1img (approved):

mykola-kobets-epam (approved):

PR #642, opened 2026-08-13
Add profiling

al1img:


src/core/sm/launcher/launcher.cpp
```diff
@@ -917,6 +923,8 @@ Error Launcher::PrepareInstance(InstanceData& instanceData)
 
 void Launcher::PrepareInstances(const Array<InstanceInfo>& startInstances)
 {
+    LOG_INF() << "[profiling] Prepare instances begin" << Log::Field("count", startInstances.Size());
```
mykola-kobets-epam, line 926: may be we can make begin end as tags too, eg
[profiling][begin] Prepare instances

mykola-kobets-epam (approved):

MykolaSuperman (approved):

src/core/sm/launcher/launcher.cpp
al1img: No need, event exporter removes [profiling] tag and send only message after tag. Adding [begin, end] tags will require changes in event-exporter and grafana dashboard.

PR #645, opened 2026-08-18
common: types: fit port ranges into the port field

MykolaSuperman:
An allowed connection may now request an inclusive port range, so a protocol with a dynamic port range (DDS/RTPS uses UDP 7400:7650) needs one rule instead of one per port:

    allowedConnections: <itemID>/<from>:<to>[/<proto>]

Exposed ports stay a flat list of individual port/protocol entries, as OCI defines them. A single port keeps working unchanged.

AOS_CONFIG_TYPES_PORT_LEN was 8, too small for "65000:65535", and is raised to 16. That also grows cExposedPortLen and cConnectionNameLen, which are only used as string capacities.


src/core/common/config.hpp
```diff
@@ -194,7 +194,7 @@
  * Port len.
  */
 #ifndef AOS_CONFIG_TYPES_PORT_LEN
-#define AOS_CONFIG_TYPES_PORT_LEN 8
```
al1img, line 197: Why it is increased? 8 chars should be enough for port.

PR #648, opened 2026-08-20
common: tools: error: construct error from both enum and errno

mlohvynenko:
This patch adds a constructor to the Error class that allows constructing an Error object from both an ErrorEnum and an errno value.
It is useful when an error is constructed from a gRPC response, for example, when the gRPC response contains both an AOS error code and an exit code (errno).
Otherwise, there is no way to propagate both error codes to the Error object, and one of them would be lost.


src/core/common/tools/error.hpp
```diff
@@ -127,21 +127,35 @@ class Error {
...
         , mErrno(errNo < 0 ? -errNo : errNo)
         , mFileName(fileName)
         , mLineNumber(lineNumber)
     {
         CopyMessage(msg);
     }
 
+    // cppcheck-suppress noExplicitConstructor
+    /**
+     * Constructs error instance from system errno value.
```
mykola-kobets-epam, line 161: please fix doxygen comments
and probably better to leave error as it used to be, ie errNo == 0 ? Enum::eNone : Enum::eRuntime

al1img (approved):

mykola-kobets-epam (approved):

PR #649, opened 2026-08-21
cm: launcher: stop instances on paused nodes

mykola-kobets-epam:


al1img (approved):

mlohvynenko (approved):

src/core/cm/launcher/nodemanager.cpp
```diff
@@ -290,7 +299,8 @@ Error NodeManager::ResendInstances(UniqueLock<Mutex>& lock, const Array<StaticSt
 
     mNodesExpectedToSendStatus.Clear();
 
-    for (auto& node : FilterActiveNodes(mNodes)) {
+    // Include paused connected nodes so running instances can still be stopped on resend.
+    for (auto& node : FilterConnectedNodes(mNodes)) {
```
MykolaSuperman: [P1] Do not resend start requests to paused nodes

Node::ResendInstances() also populates startInstances from the active instances assigned to the node; it does not check the node state. After CM restarts with persisted active instances assigned to an already paused node, receiving an empty status from that node now triggers requests to start those services again. I reproduced this through Launcher with two persisted services: the paused node receives stopInstances=0, startInstances=2. The regression test passes without this PR. Please make the paused-node resend path stop-only, regardless of persisted active assignments, rather than forwarding it to the normal start/stop resend logic.

src/core/cm/launcher/nodemanager.cpp
```diff
@@ -59,10 +70,6 @@ Error NodeManager::Start()
 
         LOG_DBG() << "Get node info" << Log::Field("nodeID", nodeID) << Log::Field("state", nodeInfo->mState);
 
-        if (nodeInfo->mState != NodeStateEnum::eProvisioned) {
-            continue;
-        }
-
         // Add online provisioned node
         mNodes.EmplaceBack();
```
MykolaSuperman: [P2] Avoid requiring monitoring for unprovisioned nodes

Removing the state filter adds unprovisioned nodes to mNodes as well. Balancer::UpdateMonitoringData() iterates this entire list and aborts balancing on the first GetAverageMonitoring() error, so an unprovisioned node that has not supplied monitoring data now prevents services from being scheduled on otherwise ready nodes. I reproduced this with three provisioned nodes with monitoring and one disconnected unprovisioned node without metrics: RunInstances() returns eNotFound. The same test passes without this PR. Please restrict which nodes are added, or update monitoring collection to skip nodes that are not relevant to scheduling, including handling missing metrics for newly tracked paused nodes.

src/core/cm/launcher/nodemanager.cpp
mykola-kobets-epam: > CM restarts with persisted active instances assigned to an already paused node
This corner case concerns also the problem of detection node status change in offline mode. There is a simple solution to that: always rebalance on start but it should be discussed on be done as a separate task.

src/core/cm/launcher/nodemanager.cpp
mykola-kobets-epam: partially fixed. Paused nodes should remain in mNodes list in order to send stop request to them, so partial mNodes filtering is dropped. But  GetAverageMonitroing can fail for such nodes, so avoid invoking a call to them.

src/core/cm/launcher/nodemanager.cpp
MykolaSuperman: I’m okay with this decision. If @al1img is okay with it as well, let’s address it in a separate ticket.

MykolaSuperman (approved):

